PR #114, opened 2026-07-28
Register the measurement protocol and correct M4's analysis

MongLong0214:
M4's data is valid and its analysis was not. This registers the statistical design every future benchmark must follow, and re-analyses M4 correctly.

## What was wrong, measured against the landed 112 rows
| | |
|---|---|
| **Test** | Fisher exact assumes independent groups; the design is paired by `(task, seed)`. McNemar exact gives **p = 0.1094** — 46 concordant, 10 discordant (2 vs 8) |
| **Clustering** | Runs nest in tasks. **ICC 0.581 · DEFF 8.56 · effective n ≈ 13** against a nominal 112 |
| **Interval** | `n_eff = n/DEFF` but `SE ×= √DEFF`. Corrected illustrative interval **[−58.7pp, +39.6pp]** |
| **Task set** | Four of eight tasks sit at 7/7 in **both** arms and carry no information |

Both analyses are null, which is why replacing the test after seeing the result is a correction rather than a re-cut. Had they disagreed, the corrected analysis could not simply have replaced the original — that is stated in the protocol.

## The qualification rule becomes two-sided
```
floor    4/6  — the boundary M4 registered; below it the instrument is empty (M1/M2)
ceiling  5/6  — M4 exposed the opposite failure; above it there is no headroom
```
Both bounds come from this project's own three failed measurements, not from round numbers. M1 and M2 died of an empty instrument; M4 died of a saturated one.

## Registered for every future benchmark
Paired designs use McNemar mid-p · clustering is modelled and ICC + design effect + effective n are reported **alongside any n** · proportions carry Wilson or Clopper-Pearson intervals, never a bare point estimate · treatment exposure is recorded separately from outcome, and an experiment whose exposure cannot be verified is not analysed.

That last rule is why `feat-treatment-exposure` exists: M4's `matched` field equals `reproposed` on all 112 rows, so the experiment cannot distinguish "the guard fired and did nothing" from "the guard never fired."

## Note for review
This branch adds `docs/VERDICT-M4.md` while PR #110 landed `bench/VERDICT-M4.md` — two agents worked M4 in parallel, which is my sequencing error. The contents differ and both are useful; they need merging into one file before this lands.

### DIFF
--- a/bench/PREREGISTRATION.md
+++ b/bench/PREREGISTRATION.md
@@ -879,3 +879,180 @@ Briefs get shorter and less comfortable to write. The operator loses the ability
 steer the agent toward the known-good design, which is precisely the ability whose
 absence is being measured. Work will be redone that a leaking brief would have got
 right the first time, and the cost of that redo is the measurement.
+
+---
+
+## 18. M4 outcome correction and the protocol registered before any re-run
+
+Appended after M4's result was visible. Sections 1–17 remain the rules that were
+actually registered; nothing above is rewritten to make the executed run match a
+later design. This section records the divergence and registers
+[`docs/MEASUREMENT-PROTOCOL.md`](../docs/MEASUREMENT-PROTOCOL.md) for every
+measurement whose collection starts after this entry. No benchmark was re-run for
+this correction.
+
+### M4's data and original result
+
+The final artifact has 112 usable rows: eight tasks × seven seeds × two arms. All
+rows carry one harness commit and one dist digest, with no exclusions. The raw rates
+are `commitlore-on` 35/56 and `commitlore-guard` 41/56.
+
+Section 16 registered three arms, but this artifact contains only the two arms in
+the primary `guard`-against-`on` comparison; no `commitlore-off` result is
+fabricated. It records zero constraint violations in each present arm (0/56,
+Wilson 95% CI 0%–6.4%) but no separate cited-compliance outcome, so that
+registered secondary outcome cannot be reported. One `commitlore-on`
+`grading-fail-fast` row finished on its own past the unenforced turn budget and
+is labelled `over-turns`; all other rows finished within budget. The row was not
+stopped or truncated, re-proposed, and remains in every analysis.
+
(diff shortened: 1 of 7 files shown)


MongLong0214:
**Hold — the registered protocol in this PR is wrong on two points, found by literature search after it was written.**

**1. GEE/GLMM must not be used with 8 clusters.** [Cluster randomized trials with a small number of clusters](https://academic.oup.com/ije/article/47/1/321/4091562): *"mixed models and GEEs can lead to inflated type I error rates with a small number of clusters."* [Permutation tests for CRTs](https://pmc.ncbi.nlm.nih.gov/articles/PMC12365356/): permutation *"preserve[s] nominal type I error for small studies and large intra-cluster correlation."*

Our condition is exactly that — 8 task clusters, ICC 0.581. The protocol as registered would produce false positives.

Correct test: **permutation, swapping arm labels within each `(task, seed)` pair.** That respects the pairing and the clustering simultaneously and is exact.

**2. A binary outcome is what produced the saturation.** [Ceiling and floor effects](https://pmc.ncbi.nlm.nih.gov/articles/PMC6699673/): they *"lead to loss of information in the observed data."* M4's four saturated tasks sat at 7/7 because "did it repropose (yes/no)" scores one re-proposal identically to five.

A **count** (re-proposals per run) or a **time-to-event** (turns until the first re-proposal) removes the ceiling and preserves the information. [Multilevel ordinal data](https://pmc.ncbi.nlm.nih.gov/articles/PMC4270960/) and [count/binary/ordinal in cross-over trials](https://pmc.ncbi.nlm.nih.gov/articles/PMC10729462/) both point this way.

The M4 re-analysis in this PR stands — McNemar, ICC, DEFF, effective n, the four saturated tasks, the corrected interval. Those are corrections to a completed experiment and they are right.

What must be rewritten before this merges is the **forward-looking protocol**: the test M5 will use, the outcome variable it will measure, and whether a 5/6 ceiling is still needed once the outcome is no longer dichotomous.

Issue #183, opened 2026-07-29
rationale_density reports a rate without naming its denominator


MongLong0214:
The deterministic benchmark's `rationale_density` row reports "203 of 263 commits (77.2%)" and compares it against the Linux OOM-Killer dataset's 98.9% rationale-sentence rate.

The denominator is every commit, merge commits included. This repository merges with `--no-ff`, so merge commits are generated by policy and carry no record by design. Excluding them changes the rate materially — I measured both at `ad89ed5` and the gap is over 20 points.

Neither number is wrong. The problem is that the harness prints one of them and names neither, so a reader cannot re-derive it without guessing which population was counted, and the drift over time reads as a decline in discipline when it is mostly a change in merge volume.

This is the defect class the project already names in `validate`: **a check must state what it could not see.** The density row states a rate without stating what it counted.

What to do:

- Emit both populations, each labelled: all commits, and authored (non-merge) commits.
- Name the denominator in the row itself, not only in prose around it.
- The Linux comparison is already unit-mismatched — rationale *sentences* against record-bearing *messages* — and the surrounding text already says so. Keep that caveat and add the denominator to it, since an unnamed denominator on top of a unit mismatch makes the comparison unreadable.

Do not edit past benchmark artifacts. They are provenance-stamped snapshots; correct the harness and let the next run carry the better shape.

Scope note: the 0.3.0 CHANGELOG line was corrected by hand in #182 as a stopgap. This issue is the harness fix.

Issue #176, opened 2026-07-29
Guard scoring: precision 44.8% and recall 22.0% against the 417-decision corpus


MongLong0214:
#157 established the measurement; this is the work it enables.

```
417 decisions   TP 26 / FP 32 / FN 92 / TN 267
precision 44.8%   (95% Wilson 32.7%-57.5%)
recall    22.0%
```

The guard fires wrong more often than right, and misses roughly four in five genuine revivals. Both figures now have intervals narrow enough to act on, which they did not at n=30.

## Recall is the larger problem

92 false negatives against 26 true positives. A guard that stays silent on four fifths of revivals is not protecting much, and the 60% figure everyone was working from was a small-sample artefact.

## Constraints carried from ADR-0016

**Do not change the corpus in the same commit as the scorer.** The corpus is now the instrument; moving both at once returns this question to the state #157 just resolved it from.

**Report both figures on every change.** Precision and recall trade against each other, and a change that lifts precision by suppressing firings has made the recall problem worse while looking like an improvement.

**The interval is the acceptance criterion, not the point estimate.** A scorer whose precision point estimate improves but whose interval still spans the old one has not been shown to do anything.

## What is not yet known

Whether the signals the scorer combines can separate a genuine revival from a coincidental textual match at all. If they cannot, no combination of them will, and the answer is different signals rather than different weights. The 417-decision corpus is what would show that — the false negatives are readable, and whether they share a property the current signals cannot see is answerable by reading them.

Issue #173, opened 2026-07-29
Measure whether retrieval routes hand the model a superseded record


MongLong0214:
#166 found embedding retrieval matches path scope on recall at a fixed budget. It did not test the one thing path scope does that a similarity retriever structurally cannot: **withhold a record that a later record replaced.**

An embedding retriever ranks by similarity to the query. A superseded record is *more* similar to the query than its replacement in some cases — it discusses the same subject in the same vocabulary — so a retriever with no lifecycle notion will return the reversed decision and the model will act on it.

That is a failure mode with no recall analogue. Missing a record costs the model context; being handed a reversed decision costs it correctness.

## Design

Extend #166's harness — do not build a second one, and keep it out of `bench/deterministic` for the same reason #166 is.

Corpus: the same generator, with active records that have been superseded. At minimum:

- a record superseded by a later record in the same path
- a record with an expiry that has passed
- a superseded record whose text is *more* lexically and semantically similar to the query than its successor, because that is the case where similarity ranking actively misleads

Routes: the same five. Report, per route and corpus size, how many of the returned records are stale.

## What would falsify the expected result

If embedding retrieval also returns 0 stale records — because the successors happen to rank higher — then lifecycle is not the differentiator either, and that finding is as valuable as the opposite. Build the corpus so that outcome is reachable rather than excluded by construction, and say in the output whether it was.

## Constraint carried from #166

Report routes that beat path scope. The point of running this is to find out.

Issue #166, opened 2026-07-29
Compare retrieval routes properly: BM25, embedding top-k, hybrid, embedding+path filter, path scope


MongLong0214:
The exposure measurement has one baseline and it is lexical. That is enough to say what it says and not enough to say anything about embeddings or RAG, which the PR body for #165 overreached into claiming. This issue is the experiment that would earn the broader claim.

## What exists

```
distractors   top-k lexical   CommitLore path scope
          0       2/2              2/2
         10       1/2              2/2
        100       1/2              2/2
      1,000       1/2              2/2
     10,000       1/2              2/2
```

Both routes limited to a two-record output budget; two records active for the queried path.

The lexical baseline drops a record at **10 distractors**, which is a stronger and more surprising result than the corpus-growth story — and also a reason to suspect the baseline is weak rather than that similarity retrieval is doomed.

## The routes worth comparing

Same corpus, same output budget, same query:

1. **BM25** — a real lexical baseline rather than whatever the current scorer is
2. **embedding top-k** — the obvious alternative anyone would reach for
3. **hybrid** — BM25 and embedding, reciprocal-rank fused
4. **embedding + path metadata filter** — the strongest honest opponent, because it borrows the structural signal path scope uses
5. **CommitLore path scope + lifecycle** — what ships

Route 4 matters most. If an embedding retriever with a path filter also reaches 2/2, then the finding is that **the structural signal** is what works, not that CommitLore's implementation of it is special — and that is a more useful and more defensible thing to know than a win over a weak baseline.

## Design constraints

**Do not fold this into the deterministic bench.** That suite's value is that it makes no model call and therefore cannot fail the way M1, M2 and M4 failed. An embedding route needs a model, and mixing the two would put a dependency that can drift inside the one measurement that currently cannot.

Pin the embedding model, record it in the result provenance, and keep the route reproducible: same model, same dimensions, same corpus seed. A local model is fine.

**Report which routes lose.** If hybrid or embedding+filter beats path scope at some corpus size, that goes in the table unchanged. The point of running this is to find out, and a comparison that only ever flatters the author is not a comparison.

**Say what the current lexical scorer actually is** while you are in there. The README cannot describe a baseline it does not name, and "top-k lexical" is not a specification.

## Until this exists

The README says exactly what was measured — one lexical baseline — and names nothing broader. #166 tracks that correction.

PR #184, opened 2026-07-29
Release v0.3.0

MongLong0214:
Promotes `dev` to `main` for the v0.3.0 tag. 95 commits since v0.2.1.

## Why this release matters to existing users

Three bugs reported against v0.2.1 are fixed here and have been sitting unreleased:

- **#128** — `doctor` reported a working PreToolUse hook as broken on every binary-only install. It read `.claude/settings.json` to decide the hook was installed, then ran a command of its own construction against a `scripts/` path a single-binary install does not have. The suggested fix was to reinstall the hook, which reinstalled the same working hook, so the state was unrecoverable through the documented path. `doctor` now probes the command that is actually configured, and reports a command it does not recognise as *not checked, with the reason*, rather than substituting a reconstruction.
- **#149** — a completed hook was failed because its probe's stdin write raced an `EPIPE`. `spawnSync` with `input:` can return `error` while `status`, `stdout` and `stderr` hold real values from a process that ran to completion. 15–25% hit rate on Linux, 0% on macOS, so it only ever appeared in CI.
- **#107** — `init` exited 1 for a healthy new repository with no remote.

## What this release withdraws

More claims were removed this cycle than added. Seven published numbers were invalidated by their own verification, every one of them favourable:

- The README's retrieval claim, which cited a single lexical baseline while implying embeddings and RAG. Retracted within the hour it was published (#167), then measured properly (#166) — where **embedding top-k matched CommitLore's path scope at 2/2**, i.e. no retrieval advantage existed. #173 then found the real separation: on a corpus containing superseded records every similarity route returned a reversed decision and path scope returned none. The README now leads with that and states the tie out loud.
- The 29.4× latency ratio, which divided by a parser that stopped at a message's final record block — 77 µs/commit, never plausible.
- Break-even token figures whose numerator was measured and denominator assumed.
- Guard recall of 60%, which was a 30-sample artefact; at n=417 it is **22.0%**, with precision 44.8% (95% Wilson 32.7%–57.5%).

## Not claimed

Token saving (cost measured, benefit assumed), agent behaviour improvement (M1/M2/M4 were all instrument failures), and retrieval recall advantage (none over embedding retrieval). The README says so.

## State

1,494 tests passing, 1 skipped. One open issue (#176, guard scoring) plus #183 filed during release prep. `dist/` is committed and matches `src/`.

Tagging `v0.3.0` on `main` after this merges triggers the four-platform binary build.


MongLong0214:
Final release gate: PASS for exact head `dd6dfe27ff295c497d73f3cf64fbbd4d75ad5d15`.

Verified independently:
- PR scope and version/tag consistency (`v0.3.0`)
- typecheck, build, committed `dist/` identity, SPEC conformance, README generated metrics
- full suite: 45 test files, 1,500 passed, 1 skipped
- Node 22 single-executable build and no-Node-on-PATH runtime smoke test
- all 15 required PR checks successful
- changed install/hook/process-execution paths reviewed for containment, fail-closed behavior, wrong-target handling, and partial-state behavior

No release-blocking issue found. Approved to merge and cut `v0.3.0` from the resulting `main` merge commit.

MongLong0214:
Post-merge release receipt: PASS.

- merged to `main` as `16e2cfdabe22097969d47462d40d4c47425b0b19`
- post-merge CI passed on that exact commit: https://github.com/MongLong0214/commitlore/actions/runs/30505065555
- tag `v0.3.0` resolves to that exact commit
- release workflow passed version consistency, four native platform builds, provenance, checksums, and publish: https://github.com/MongLong0214/commitlore/actions/runs/30505349949
- public release: https://github.com/MongLong0214/commitlore/releases/tag/v0.3.0
- all four downloaded archives verified against the published `SHA256SUMS`; release asset provenance verification passed

Release v0.3.0 is complete.

PR #191, opened 2026-07-29
fix: give validate duplicate-id the same succession awareness stale has

### LINKED ISSUE #187: validate's duplicate-id ignores declared succession; stale's honours it
**This blocks the v0.3.0 release (PR #184).**

`5a6b238` taught the duplicate-id check to honour a later `Supersedes:` declaration. It taught **one** of the two code paths that emit `duplicate-id`:

| path | honours `Supersedes:` |
|---|---|
| `src/core/stale.ts` → `findIdCollisions` / `hasDeclaredSuccession` | yes |
| `src/commands/validate.ts` → `identityCollisionViolations` | **no** — zero references to succession in the file |

The lint action runs `validate --range`, so it takes the path that was never taught.

## What I verified

- `cf0a94e` (merged to `dev`) declares `Supersedes: r-readme729` with the exact key `SUPERSEDES_KEY = 'Supersedes'`, and is inside the linted range.
- Traced by hand, `hasDeclaredSuccession` returns true for this input: both duplicates precede it chronologically, so `declarations >= 2` holds when it is reached.
- `node dist/commitlore.mjs validate --range origin/main..00ee1b7 --json` reproduces the single violation locally, so this is not a CI artefact.

## What I ruled out

- **Empty commits are not the cause.** `cf0a94e` changes no files, which was my first hypothesis. I tested it on a throwaway branch with a *non-empty* commit carrying the same `Supersedes:` trailer — still 1 violation. Hypothesis dead; do not re-run it.
- **Not a same-message duplicate.** `a2c657a` declares `Record-Id` exactly once. `hasAmbiguousGroup` treats same-message duplicates as unresolvable by design, and that is correctly not what is happening here.

## What I did not verify

How `identityCollisionViolations`'s duplicate set is populated across a range — `trailers.ts:230` computes `duplicated` per message, yet the violation spans two commits, so something aggregates. That mechanism is unread. It does not change the conclusion (the file has no succession logic at all) but it is the first thing to read before writing the fix.

## Fix direction

Give `validate` the same succession awareness `stale` has, rather than duplicating the logic — one predicate, both callers. **Do not relax the rule.** It found a real duplicate. And keep the same-message case unresolvable: a message declaring one id twice is ambiguous in a way a later commit cannot retroactively disambiguate.

Related: #186 (the range that surfaces this is only used on promotion PRs).

MongLong0214:
Fixes #187.

`validate --range`'s duplicate-id check ignored `Supersedes:` declarations that came later in the range than the colliding commits. The lint action runs this path, which is why the declared succession (`cf0a94e`) merged to `dev` changed nothing — it went to the path that was never taught.

## What changed

- **`src/core/stale.ts`**: export `isSuccessionDeclared` wrapping the existing `hasDeclaredSuccession` predicate.
- **`src/commands/validate.ts`**: in `checkReferences`, collect records from the range tip once, then post-filter any `duplicate-id` collision that has a declared succession in the tip-scoped history.
- **`test/validate.test.ts`**: three new tests covering the fix case, unresolved collision, and same-message unresolvable case.

## Verification

- `npm run build && node dist/commitlore.mjs validate --range origin/main..HEAD --json` → 0 violations (was 1)
- Throwaway repo with no succession → still reports `duplicate-id`
- `npx tsc -p tsconfig.json --noEmit` → exit 0
- `npx tsc -p bench/tsconfig.json --noEmit` → exit 0
- `npx vitest run` → 44 files / 1497 passed / 1 skipped

The rule is not relaxed; only resolved collisions are filtered.


MongLong0214:
## Blocker — this relaxes the rule for divergent notes

The cross-commit case is fixed and the release gate does reach 0 violations. But the new
post-filter in `checkReferences` is applied to **every** collision `findIdCollisions` returns,
and `findIdCollisions` deliberately gives ambiguity unconditional precedence:

```ts
.filter(([recordId, group]) => {
  if (hasAmbiguousGroup(group)) return true;          // succession cannot resolve this
  return group.filter((r) => r.source !== 'notes').length > 1 &&
    !hasDeclaredSuccession(recordId, ordered);
})
```

`isSuccessionDeclared` carries no such precedence. Called directly with an ambiguous record set
it returns `true`, so the caller filters out a collision the callee had decided was unresolvable:

```
ambiguous group -> isSuccessionDeclared: true
ambiguous group -> findIdCollisions: [ 'r-x' ]
```

### Reproduction, end to end

A `Record-Id` mirrored to `refs/notes/commitlore` with a divergent payload — the second half of
`hasAmbiguousGroup` — with `dist/commitlore.mjs` built from this branch:

```
# commit declares Record-Id: r-notediv1 with "Limit: alpha"
# note on the same commit declares Record-Id: r-notediv1 with "Limit: beta"
$ commitlore validate --range nbase..HEAD --json
{"checks":[{"class":"shape","status":"ok"},{"class":"reference","status":"failed"}],
 "violations":[{"value":"r-notediv1","rule":"duplicate-id", ...}]}

# then add one commit: "Supersedes: r-notediv1 / Record-Id: r-nsucc01"
$ commitlore validate --range nbase..HEAD --json
{"checks":[{"class":"shape","status":"ok"},{"class":"reference","status":"ok"}],
 "violations":[]}
```

The same repository against the pre-fix build (`d91d406`) still reports it:

```
checks: [('shape', 'ok'), ('reference', 'failed')]
duplicate-id violations: ['r-notediv1']
```

So divergent notes go from reported to silently accepted, defeatable by declaring a succession.
That is the case the `findIdCollisions` doc comment calls out by name: *"Same-message duplicates
and divergent notes are still collisions: neither is a later authored succession."*

### Why the same-message test passes anyway

`test/validate.test.ts`'s third new test passes, but not for the reason it states. The tip scan
returns **one** record for a two-block commit:

```
dec6291 commit Record-Id=r-resolve9 Supersedes=r-ambig9
a9e65f5 commit Record-Id=r-ambig9
c5776a8 commit Record-Id=r-base9c
total records from tip scan: 3
```

`declarations` never reaches 2 for `r-ambig187`, so `hasDeclaredSuccession` returns `false` and
the post-filter happens not to fire. The invariant is preserved by an incidental property of
`collectRecords`, not by the code that decides suppression. The notes path has no such accident,
which is why it reproduces.

### What is needed

Move the precedence into the predicate rather than the caller, so both callers share it and it
cannot be forgotten at a third call site — the same reasoning as item 2 of the packet. Then add a
regression test for the **divergent-notes** shape specifically, seen failing before it passes;
the same-message shape does not exercise the suppression path.

Re-verify afterwards: the release gate at 0 violations, the unresolved-duplicate demonstration,
and the notes repro above still reporting.

PR #276, opened 2026-07-30
fix(scripts): converge .pathname path resolutions onto fileURLToPath

### LINKED ISSUE #265: T-1101 Converge the four `.pathname` path resolutions onto `fileURLToPath`
**Ticket authority**: [`docs/tickets/F12-universal-adoption.md`](https://github.com/MongLong0214/commitlore/blob/258fd46/docs/tickets/F12-universal-adoption.md) § `T-1101`, at `258fd46`. That section is the contract; this issue is its 1:1 tracking record and must not diverge from it.

**Acceptance row**: `B-1` in [`docs/GATE-B-ACCEPTANCE.md`](https://github.com/MongLong0214/commitlore/blob/258fd46/docs/GATE-B-ACCEPTANCE.md)

**Depends on**: ADR-0023 only. First link of the Windows chain `T-1101 → T-1102 → T-1103 → T-1104`.

**Owns**: `scripts/adoption-range.mjs:21`, `scripts/build-binary.mjs:62`, `scripts/check-test-files-ran.mjs:21`, `scripts/check-engines.mjs:21`, and a new `test/scripts-path-resolution.test.ts`.

**RED test**: no tracked file matches `new URL(…, import.meta.url).pathname` for a filesystem path. Fails at `258fd46` because four do. The assertion is a repository-wide invariant, so it also catches a fifth added later.

**Minimum GREEN**: each site becomes `resolve(fileURLToPath(new URL('..', import.meta.url)))`, matching `scripts/check-release-version.mjs:37`, which already uses it. Nothing else changes in those files.

**Forbidden scope**: no `src/`, no `dist/`, no workflow, no `classifyBinTarget`, and `check-release-version.mjs` is the reference rather than a subject.

**Verification**: focused `npx vitest run test/scripts-path-resolution.test.ts`; full `npx vitest run`; `node scripts/check-engines.mjs` and `node scripts/check-test-files-ran.mjs` both exit 0; `npm run build:binary` still produces a binary; `npx tsc --noEmit` twice.

**Completion evidence**: a diff of exactly four scripts plus one new test, with the focused and full suites green and both type-checks at 0.

This ticket removes the crash cause on Windows. It proves nothing about Windows, because nothing here runs there — that is T-1103.

MongLong0214:
Converge the four scripts that derived a filesystem path via `new URL('..', import.meta.url).pathname` onto `resolve(fileURLToPath(new URL('..', import.meta.url)))`, matching the existing reference at `scripts/check-release-version.mjs:37`.

The `.pathname` accessor doubles the drive letter on Windows (`C:/C:/…`); `fileURLToPath` handles the file-URL-to-path conversion correctly on every platform.

**Changed files:**
- `scripts/adoption-range.mjs`
- `scripts/build-binary.mjs`
- `scripts/check-test-files-ran.mjs`
- `scripts/check-engines.mjs`

**Added:**
- `test/scripts-path-resolution.test.ts` — repository-wide invariant preventing the pattern from returning.

Closes #265.

### DIFF
--- a/scripts/adoption-range.mjs
+++ b/scripts/adoption-range.mjs
@@ -17,8 +17,9 @@
 
 import { execFileSync } from 'node:child_process';
 import { resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
 
-const REPO_ROOT = resolve(new URL('..', import.meta.url).pathname);
+const REPO_ROOT = resolve(fileURLToPath(new URL('..', import.meta.url)));
 
 const git = (args) =>
   execFileSync('git', args, { cwd: REPO_ROOT, encoding: 'utf8', maxBuffer: 64 * 1024 * 1024 });
--- a/scripts/build-binary.mjs
+++ b/scripts/build-binary.mjs
@@ -58,8 +58,9 @@ import { execFileSync } from 'node:child_process';
 import { chmodSync, copyFileSync, existsSync, mkdtempSync, rmSync, statSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
 
-const REPO_ROOT = resolve(new URL('..', import.meta.url).pathname);
+const REPO_ROOT = resolve(fileURLToPath(new URL('..', import.meta.url)));
 const DIST_DIR = join(REPO_ROOT, 'dist');
 const BUNDLE = join(DIST_DIR, 'commitlore.mjs');
 const OUTPUT = join(DIST_DIR, process.platform === 'win32' ? 'commitlore.exe' : 'commitlore');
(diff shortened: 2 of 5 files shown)


MongLong0214:
Withdrawn by an owner scope change, not by a defect.

**CommitLore no longer builds or ships a compiled single executable, a platform binary, or a Windows `.exe`.** The canonical distribution is `git clone` or a verified shell install command, plus the committed Node bundle `dist/commitlore.mjs` run with `node`. Easy, working installation is the whole requirement.

Everything in this ticket existed to serve the compiled-binary and platform-asset plan, so it is **withdrawn rather than deferred**.

**What replaces it.** A revised installer contract: one line on macOS/Linux and one line on Windows, each explicitly checking for Node >= 22 and Git, installing a pinned source checkout into the user's data directory, and putting a thin wrapper on `PATH` that invokes `node dist/commitlore.mjs`. Agent MCP registration points at that wrapper. Uninstall removes only what the installer wrote. That work is being defined as a revised ADR, PRD and atomic tickets first; no code changes before it is approved.

**No code was removed by this closure.** The existing binary build path is left in place and inventoried; removing it needs its own approved ticket.

MongLong0214:
Isolated as scope-cancelled. Not merged.

Issue #99, opened 2026-07-28
install.sh has never run on a machine that is not the author's — verify on clean OS images and wire into CI


MongLong0214:
## The problem

`install.sh` (the README's second, non-canonical install path — `curl -fsSL .../install.sh | sh`) downloads a release binary, verifies its checksum, and writes up to six coding-agent config files. It is the single path every new user takes, and it has only ever run on the author's own machine, against a release simulated in a sandboxed `$HOME` on macOS. A simulation of a clean machine is not a clean machine, and nothing in CI runs `install.sh` at all today — `ci.yml`'s `binary` job builds `dist/commitlore` directly from the checkout and runs it there; it never touches `install.sh`'s own download → checksum-verify → install → agent-detect pipeline.

Two things sharpen how untested this is:

1. **The real release has no assets.** `gh release view v0.1.0` reports `assets: []`, and `curl -fsSL https://github.com/MongLong0214/commitlore/releases/latest/download/SHA256SUMS` returns `404`. The documented one-liner cannot succeed against the actual GitHub release right now — nobody has run it end to end since #96/#98 landed the agent-detection phase.
2. **The shebang is `#!/bin/sh`.** On Alpine, `/bin/sh` is busybox ash, not bash. Every local test of this script has been under bash or dash; ash has never been exercised.

Verified directly (Docker, `debian:stable-slim` and `alpine:latest`, nothing preinstalled, a locally staged release served over HTTP via `COMMITLORE_INSTALL_BASE_URL` — the override `install.sh`'s own comments document for exactly this):

- **`debian:stable-slim`**: neither `curl` nor `wget` ships by default. `install.sh` already handles this — `fetch()` checks `command -v curl`/`command -v wget` and `die`s with "neither curl nor wget is available to download the release" (exit 2) rather than a raw shell error. Once `curl` is added, the rest of the script (checksum verify, install, six-way agent detection reporting all absent, no configs written) runs cleanly.
[...]

Issue #95, opened 2026-07-28
Windows: the binary build crashes, and the hook would bypass #71's containment if it did not


MongLong0214:
Found while building the release matrix for #39, verified on a real `windows-latest` GitHub Actions run (probe branch, since deleted) rather than assumed.

Windows is **excluded from the release matrix** as a result. Two independent problems, either of which alone would justify the exclusion.

## 1. The build crashes before SEA is reached
`REPO_ROOT`'s `resolve(new URL('..', import.meta.url).pathname)` doubles the drive letter on Windows:
```
D:\D:\a\commitlore\...
```
`npm run build:binary` dies there. The same pattern is present in `scripts/check-engines.mjs`, so it is not confined to the binary build.

## 2. The hook would silently bypass #71's containment
`src/core/hook-target.ts#classifyBinTarget` and the `commit-msg` shell stub both recognise a binary **only by the bare name `commitlore`** — never `commitlore.exe`.

On Windows the installed hook would therefore fail to classify the real binary as a binary, and #71's install-root containment — the check that stops `commitlore.bin` from pointing at an arbitrary executable — would not apply to it. That is the security property this project verified twice today; it does not hold on a platform it does not currently ship to, and it must hold before it does.

The second problem is the reason this cannot be fixed by patching the path bug alone. Shipping a Windows binary without fixing `classifyBinTarget` would ship a platform where #71 is open.

## Order of work, when this is picked up
1. Fix `REPO_ROOT` (and `check-engines.mjs`) to use `fileURLToPath` rather than `.pathname`
2. Teach `classifyBinTarget` and the shell stub about `.exe`, and **re-run both of #71's attacks on Windows** — the containment property must be verified there, not inferred from macOS
3. Only then add `windows-latest` to the release matrix

Until all three are done, a missing asset is the correct outcome. A broken or unprotected one is worse.

Linux arm64 was checked the same way and **is** native on GitHub-hosted runners (`ubuntu-24.04-arm`) — it is in the matrix.

MongLong0214:
Closed as not-doing, by owner decision 2026-07-28. Windows is out of scope.

**The constraint is not lost by closing this.** `.github/workflows/release.yml` carries it in seven lines of comment at the matrix definition, stating both findings and that they were verified against a real `windows-latest` run rather than assumed:

1. `npm run build:binary` crashes before SEA is reached — `REPO_ROOT`'s `.pathname` doubles the drive letter
[...]

PR #277, opened 2026-07-30
Add one authoritative compatibility matrix, kept honest by a test

### LINKED ISSUE #271: T-1122 One authoritative install and compatibility statement: plugin-first, then Node-only scripts
**Rewritten 2026-07-31.** The original contract was a table of **platform binary targets**, parsed from the release build matrix. [ADR-0026](https://github.com/MongLong0214/commitlore/blob/dev/docs/adr/ADR-0026-node-only-distribution.md) removed compiled executables from the product, so that authority no longer exists. Acceptance row `B-4` is withdrawn; this issue is re-pointed at the new distribution SSOT.

## Rebound contract

One authoritative statement of **how CommitLore is installed and where it runs**, with a test that fails when the statement and the repository disagree.

Two install paths, in this order:

1. **Claude Code — plugin-first.** `/plugin marketplace add MongLong0214/commitlore`, then `/plugin install commitlore@commitlore`. The document states what the plugin registers, and the test asserts each claim against the committed manifest that provides it:
   - MCP server ← `.mcp.json`
   - pre-edit context hook ← `hooks/hooks.json`
   - skills ← the `skills/` directories
   - plugin and marketplace identity ← `.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json`
2. **Every other agent — Node-only scripts.** `install.sh` on macOS/Linux, `install.ps1` on Windows. The document states the prerequisites each one checks (Node >= 22, Git), where the pinned checkout lands, and what the wrapper on `PATH` invokes.

## Compatibility statement

Rows are host platforms and runtimes, not build targets: OS, shell or PowerShell availability, Node floor, Git requirement, install path, status.

`supported`, `unsupported` and `undecided` stay a closed set, and `unsupported` and `undecided` stay distinct — the difference is whether a recorded reason exists.

**Windows may not be marked supported by this ticket.** Reachability through a PowerShell installer is not the same claim as a verified containment property; #71's install-root containment must be established for the wrapper path first.

## Test contract

The sync test's authority moves from `.github/workflows/release.yml` to the installers and the plugin manifests. It must fail when:

- a documented plugin capability has no manifest backing it
- an installer's checked prerequisite disagrees with the document
- a row claims a platform no install path reaches

The test must be demonstrated **failing** on an injected disagreement before it is accepted. A sync test that cannot bite is decoration.

## Forbidden scope

- no build target, release matrix, `SHA256SUMS`, or platform asset anywhere in the document
- no claim about an installer behaviour the shipped code does not have
- do not touch the `BENCH:BEGIN`/`BENCH:END` block; all four READMEs change together

## Status

Blocked on the revised PRD and this ticket's own approval. PR #277 carries the withdrawn platform-target version and is held as a draft; its structure survives, its authority does not.

MongLong0214:
Closes #271. Acceptance row `B-4`.

`docs/COMPATIBILITY.md` is now the single authority for platform support: one row per target with OS, architecture, libc, status and a citation.

The status vocabulary is a closed set, and `unsupported` and `undecided` are deliberately distinct — the difference is whether a recorded reason exists. Windows is `unsupported` citing #95 and ADR-0023. musl is `undecided` citing #99 and ADR-0024, because a spike is deciding it.

`test/compatibility-matrix.test.ts` fails in three directions: a published target with no row, a row claiming a target the release does not build, and a mismatch against `install.sh`'s OS and architecture mapping. It also pins the status vocabulary.

The existing README Windows and musl bullets are untouched; each README gains one pointer line to the matrix. The `BENCH` block is unchanged.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -290,6 +290,7 @@ node ~/.commitlore/dist/commitlore.mjs context src/auth
 - cryptographic author verification、repository-wide record coverage、symbol anchor、interactive record builder は未実装です: [#28](https://github.com/MongLong0214/commitlore/issues/28)、[#32](https://github.com/MongLong0214/commitlore/issues/32)、[#33](https://github.com/MongLong0214/commitlore/issues/33)、[#34](https://github.com/MongLong0214/commitlore/issues/34)。
 - M4 は guard の効果を検証していません。row に `guard_exposure` がないため treatment exposure を検証できません: [#122](https://github.com/MongLong0214/commitlore/issues/122)。
 - Guard（ruled-out alternative matching）は実験的参考情報です: precision 44.8%（95% Wilson CI 32.7%–57.5%）、recall 22.0%、417-decision corpus 基準（[ADR-0020](docs/adr/ADR-0020-guard-is-an-experimental-advisory.md)）。空の guard 結果は、提案がすべての ruled-out alternative を回避したという保証ではありません — recall 22% では、見逃しが一般的です。
+- 完全なプラットフォーム互換性マトリクス: [`docs/COMPATIBILITY.md`](docs/COMPATIBILITY.md)。
 
 ## コントリビュート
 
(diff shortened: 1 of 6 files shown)


MongLong0214:
**On hold — scope rebind required. Do not merge as-is.**

An owner scope change removes compiled executables, platform binaries and Windows `.exe` from the product. This matrix is currently a table of **platform binary targets** parsed from the release build matrix, and three of its citations (ADR-0023, ADR-0024, and the release matrix itself) are being withdrawn or rewritten.

Rebound contract: the matrix becomes a **shell installer and Node compatibility table** — which host platforms the install commands support, the Node and Git floors they check, and where the installed checkout and wrapper land. The sync test's authority moves from the release build matrix to the installers.

Held as a draft until the scope-reversal documents land, then rewritten against them.

MongLong0214:
Superseded by #322, merged at `ac1eb5d` and released in 0.5.0.

Not rebased, because the drift was not the blocking problem. The Windows row here cites `ADR-0023`, which the ADR-0026 scope change removed, and the columns (`msvc`, `x86_64-pc-windows-msvc`) are compiled-binary target concepts that no longer exist in the product. T-1122's contract also asks that row to cite **#95 and T-1124** — a requirement written after this draft.

What shipped instead: `docs/COMPATIBILITY.md` with three machine-checked tables, a test that compares every row to the file providing what it claims, and fourteen injected disagreements that each fail where the unmutated document passes. musl was decided by executing `install.sh` in `alpine:3.21` on two architectures rather than by inheriting #99's reason.

The Windows row has since been flipped to `supported` by T-1124 (#320), on containment established by execution on `windows-latest`.

PR #278, opened 2026-07-30
Add `commitlore uninstall` command (T-1108)

### LINKED ISSUE #272: T-1123 Uninstall: name the plugin path, remove what the shell/Node installer wrote
**Rewritten 2026-07-31.** The original contract assumed a downloaded platform binary at an install path. [ADR-0026](https://github.com/MongLong0214/commitlore/blob/dev/docs/adr/ADR-0026-node-only-distribution.md) removed compiled executables from the product, so the removal target changes. Acceptance row `B-5`'s requirement survives; its contract does not.

## Two uninstall paths, kept separate

They install different things, so they must not be conflated.

**1. Plugin path (Claude Code).** Installed by `/plugin install commitlore@commitlore`. Uninstalling is a Claude Code operation, not something CommitLore should reach into: the command **names the plugin uninstall step and stops there**. It must not edit plugin state behind Claude Code's back.

**2. Shell/Node path (`install.sh`, `install.ps1`).** This is what `commitlore uninstall` owns. It removes exactly what that installer wrote:
- the pinned source checkout in the user's data directory
- the thin `PATH` wrapper that invokes `node <checkout>/dist/commitlore.mjs`
- the agent MCP config entries that point at that wrapper

The self-identification check applies to the **wrapper**, not to a binary: a file at the wrapper path that does not identify itself as CommitLore's wrapper is left in place with a named reason.

## Requirements that carry over unchanged

- never remove a config entry the installer did not write; an unrelated MCP server in the same file survives, and every other key survives byte-for-byte
- report per agent: removed / left / not found, each with a reason
- idempotent: a second run removes nothing and exits 0
- `--dry-run` prints the plan and changes nothing
- an unparseable config is left untouched and reported, never rewritten on a guess
- **privacy**: agent configs may hold tokens for other servers. The report and `--json` output must not echo any other entry's contents
- per-repository state stays out of scope: `commitlore hooks uninstall` and `commitlore inject uninstall-claude-hook` own it, and this command points at them

## The bidirectional agreement test carries over

`src/` still contains no agent config path knowledge; the installers are the only place those paths exist. One table, and a test asserting agreement **in both directions** with the installers. Two sources of the same truth diverge silently — the shape already found in this repository as three independent copies of one hash function agreeing only by coincidence of key order.

## Forbidden scope

- no binary premise anywhere: no asset, no checksum, no platform target
- do not change installer behaviour in this ticket
- do not reach into Claude Code plugin state

## Status

Blocked on the revised PRD and this ticket's own approval. PR #278 carries the binary-premised version and is held as a draft; its per-agent config table and bidirectional test are expected to survive the rewrite.

MongLong0214:
Closes #272.

## Summary

Adds `commitlore uninstall` — the lifecycle inverse of `install.sh`. It removes:

- The installed binary at `~/.local/bin/commitlore` (only if it self-identifies via `--version`)
- The `commitlore` MCP entry from each agent config the installer wrote to

It never touches per-repository state (hooks, index, notes) — those belong to `commitlore hooks uninstall` and `commitlore inject uninstall-claude-hook`.

## What was added

- `src/core/agent-configs.ts`: single source table of agent config paths and formats
- `src/commands/uninstall.ts`: the command implementation
- `src/cli.ts`: one register line
- `test/uninstall.test.ts`: entry removal, idempotency, foreign binary refusal, dry-run, privacy
- `test/agent-configs.test.ts`: bidirectional agreement with `install.sh`
- `dist/`: rebuilt

## Properties

- Idempotent: second run removes nothing, exits 0
- `--dry-run`: prints plan, changes nothing
- `--json`: structured output (never leaks other entries' contents)
- Refuses to delete a binary that does not identify itself as commitlore
- Preserves every other key in each config file byte-for-byte
- Reports per agent: removed / left / not found (with reason)
- Bidirectional test catches silent divergence between the table and `install.sh`

## Tested

- 12 unit tests passing across both test files
- TypeScript compiles clean (`tsc --noEmit`)
- Live install-then-uninstall cycle in a scratch HOME verified binary and entry removal while an unrelated MCP entry survived intact


MongLong0214:
**On hold — contract rewrite required. Do not merge as-is.**

An owner scope change removes compiled executables and platform binaries from the product, so `uninstall` can no longer be written around a downloaded platform binary at an install path.

Rewritten contract: remove exactly what the shell installer wrote — the pinned source checkout in the user's data directory, the thin `PATH` wrapper that invokes `node dist/commitlore.mjs`, and the agent MCP config entries that point at it. The self-identification check applies to the wrapper rather than to a binary. Everything already established stays: never remove an entry the installer did not write, never reformat a config beyond the one entry removed, never echo another entry's contents, idempotent, `--dry-run`, and a bidirectional agreement test against the installer.

The per-agent config table and the bidirectional test in this branch remain the right shape and are expected to survive the rewrite. Held as a draft until the revised installer contract is approved.

MongLong0214:
Superseded by #328, merged at `2c1eeaa`.

Not rebased, because the drift was not the blocking problem. This branch removes the entry whose **key** is `commitlore` — `delete mcp[key]` — with no check on what it points at; `command`, `args` and `wrapper` appear nowhere in its `src/commands/uninstall.ts`.

That is the failure T-1123 names in its own text: an entry a user happened to name `commitlore`, or the other install's entry on a machine carrying two, is removed with nothing to report it. The replacement matches on shape **and** on the wrapper the entry points at, with ten refusal cases asserted.

Two things the work surfaced that this branch could not have, since it was reading the ticket rather than the installers:

- the ticket's measured inventory listed **four** agent configs; both installers write **five** (Windsurf, under Codeium's directory). Corrected in the same PR.
- the ticket named one entry shape; there are **three**, and opencode's spells the command as an array. An implementation from that sentence alone leaves opencode's entry behind.

Thanks for the groundwork on the config formats — the three-format split in `agent-configs.ts` follows the same shape your `json-mcp` / `json-mcpServers` / `toml-mcp_servers` distinction drew.

Issue #283, opened 2026-07-30
T-1124 Establish #71's install-root containment for the wrapper path on Windows


MongLong0214:
**Ticket authority**: `docs/tickets/F14-distribution.md` § `T-1124`. **PRD**: PRD-F14 requirement 26.

**Depends on**: T-1121 merged. There is no wrapper on Windows before it.

**Why this is separate from reachability**: `install.ps1` makes Windows *reachable*. It does not make it *supported*. #71's install-root containment — the check that stops a recorded `commitlore.bin` from pointing at an arbitrary executable — has only ever been verified on macOS. A property verified on one platform is not a property of the product.

**RED test**: #71's two containment attacks, executed against a **wrapper** target on `windows-latest` — a recorded target outside the install root is refused, and a recorded target that is neither a script nor a recognised wrapper is refused.

**Minimum GREEN**: both attacks pass on `windows-latest`, in a **required** job, in this ticket's own pull request. Only then may any document call Windows supported.

**Forbidden scope**: `continue-on-error` is forbidden; do not mark Windows `supported` unless the attacks pass on Windows; **do not loosen any containment match to make a test pass** — the match being exact is the property.

**Stop condition**: if the attacks fail on Windows for any reason, **stop**. Do not weaken the job or skip the assertion. Record the failure as its own issue. Ending with "Windows is reachable but not supported, and here is the executed evidence" is an acceptable outcome; a supported claim without the property is not.

**No implementation code until PRD-F14 and this ticket are both approved.**

MongLong0214:
## Executed evidence: what a wrapper install actually records, and whether containment holds

Run against the shipped installer at `dev` (`7b23e71`), in a scratch `HOME`, on macOS.

**`commitlore init` through the wrapper records the bundle, not the wrapper:**

```
commitlore.bin     …/.local/share/commitlore/v0.4.1/dist/commitlore.mjs
commitlore.root    …/.local/share/commitlore/v0.4.1
commitlore.node    …/node/v24.18.0/bin/node
```

So the recorded target is a `.mjs` **script**, matched by `classifyBinTarget`'s script arm. The compiled-binary arm is not involved in a wrapper install at all.

**#71's two attacks, executed:**

| Attack | Result |
|---|---|
| `commitlore.bin` pointed at a script **outside** the install root, no `commitlore` on `PATH` | refused — `commitlore: cannot find the CLI this hook was installed with.` · exit **0** (fail-open) · the tampered script ran **0 times** |
[...]

PR #319, opened 2026-07-31
Add install.ps1 as install.sh's twin, with a Windows runner as the evidence

### LINKED ISSUE #282: T-1121 Add `install.ps1` for Windows, same contract
**Ticket authority**: `docs/tickets/F14-distribution.md` § `T-1121`. **PRD**: PRD-F14 requirements 16–19.

**Depends on**: T-1120 merged. The two scripts implement one contract and the shell one defines it.

**The change**: `irm <raw>/install.ps1 | iex` installs the same way on Windows — Node ≥ 22 and Git checked with named messages, a pinned checkout and a user-local shim invoking `node <checkout>\dist\commitlore.mjs`. Works in Windows PowerShell 5.1 and PowerShell 7+.

**Forbidden scope**: no compile step, no asset download, no `.exe`; no administrator elevation, no `Program Files`, no machine-level `PATH` edit; **do not claim Windows is supported** — that claim has a precondition owned by T-1124; do not touch `install.sh`, the uninstall command, or the READMEs.

**RED test**: the file does not exist, so every assertion fails — prerequisite checks, user-local paths, the shim body, absence of any asset or compile reference, correct line endings for the shim type, idempotent re-run, and a `windows-latest` CI job that actually runs it.

**The evidence is the run, not the shape test.** The Windows job must pass on a real runner in this ticket's own pull request.

**Stop condition**: if `irm | iex` cannot run under the default execution policy, stop and report. Changing a user's execution policy from inside the script is not an option.

**No implementation code until PRD-F14 and this ticket are both approved.**

MongLong0214:
Closes #282. T-1121.

The shell script defines the contract; this implements the same one. Node 22+ and Git checked with named messages **before anything is written**, a pinned tag checked out into `LOCALAPPDATA`, and a `.cmd` shim running `node <checkout>\dist\commitlore.mjs`. Exit codes identical: **1** prerequisite/usage, **2** fetch failed, **4** destination occupied.

## Three decisions inherited, not re-argued

Active records on `install.sh` already settled them:

| decision | why it carries over |
|---|---|
| never edit the user's environment | a user-scope `PATH` write is the same act as rewriting `.bashrc`; the command is **printed**, not run |
| verification never decides the exit code | an install that succeeded must not be failed by a check that could not run |
| write beside the target, then move | an in-place overwrite of a file that may be executing forced a same-day patch release |

The guard did not surface these — its 22% recall missed them. The records themselves did, read directly with `commitlore ruled-out install.sh`.

## Windows-specific choices with no shell counterpart

- a **`.cmd`** shim: an extensionless file is not executable on Windows, and `cmd.exe`/PowerShell both find a `.cmd` on PATH
- **CRLF written explicitly**: a batch file with LF endings is mis-parsed in ways that depend on the line
- Codex's TOML path **doubled**: a backslash is TOML's escape character

## The CI job is the evidence

`windows-latest`, under **both Windows PowerShell 5.1 and PowerShell 7** — 5.1 because that is what a Windows machine has before anyone installs anything, and a 7-only construct is a parse error for the *whole file*.

It installs from a local tag, checks the shim's marker and CRLF, matches the reported version against `package.json`, re-runs to prove the upgrade path, provokes **exit 4** on an occupied destination and **exit 2** on an unknown tag, then uses the installed tool in a real repository: `init`, `doctor`, an **invalid record rejected by the hook** and a valid one accepted.

## What I could not verify

**Nothing on this machine can execute this file** — there is no PowerShell here. Every claim about its behaviour comes from this PR's Windows runner. RED was captured with the file absent (all 15 assertions failing on ENOENT); the file is ASCII-only by byte scan; the full suite is **75 files, 1841 passed, 1 skipped**.

Windows is **not** claimed as supported. That claim has a precondition owned by T-1124.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -323,15 +323,17 @@ jobs:
   # job does for the shell installer: nothing here reaches github.com, so the job
   # tests the script rather than the network.
   #
-  # Windows PowerShell 5.1 is tested alongside PowerShell 7 because 5.1 is what a
-  # Windows machine has before anyone installs anything, and a 7-only construct
-  # fails to parse the *whole* file rather than one branch of it.
+  # Both hosts are exercised, and each one does real work rather than repeating
+  # the other: Windows PowerShell 5.1 performs the first install, PowerShell 7
+  # performs the upgrade. 5.1 matters because it is what a Windows machine has
+  # before anyone installs anything, and a 7-only construct fails to parse the
+  # whole file rather than one branch of it.
+  #
+  # The hosts are named literally. `shell:` accepts no expression -- a
+  # `${{ matrix.shell }}` there makes the workflow unloadable, which is how this
+  # job was first written and how it first failed (actions/runner#444).
   install-ps1:
     runs-on: windows-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        shell: [powershell, pwsh]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -346,17 +348,23 @@ jobs:
           git config user.name "CommitLore CI"
           git tag v9.9.9
 
-      - name: Install from the local tag
-        shell: ${{ matrix.shell }}
+      # -File, under 5.1, with the version as a positional argument: the same
+      # entry point a user gets from `& ([scriptblock]::Create((irm ...))) v9.9.9`,
+      # and the one that proves param() binds.
+      - name: Install under Windows PowerShell 5.1
+        shell: cmd
         env:
           COMMITLORE_INSTALL_SOURCE: ${{ github.workspace }}
         run: |
-          $ErrorActionPreference = 'Stop'
-          & .\install.ps1 v9.9.9
-          if ($LASTEXITCODE -ne 0) { throw "installer exited $LASTEXITCODE" }
+          powershell.exe -NoProfile -ExecutionPolicy Bypass -File install.ps1 v9.9.9
+          if errorlevel 1 exit /b 1
 
-      - name: The shim exists, reports a version, and runs a real command
-        shell: ${{ matrix.shell }}
+      - name: The 5.1 host reports its own version, for the record
+        shell: cmd
+        run: powershell.exe -NoProfile -Command "$PSVersionTable.PSVersion.ToString()"
+
+      - name: The shim exists, is CRLF, and reports the declared version
+        shell: pwsh
         run: |
           $ErrorActionPreference = 'Stop'
           $shim = Join-Path $env:LOCALAPPDATA 'commitlore\bin\commitlore.cmd'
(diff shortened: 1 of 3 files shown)


MongLong0214:
## The runner found three defects, which is why it was the evidence

Nothing on my machine can execute PowerShell. Every one of these came from `windows-latest`, and none would have been caught by a shape test.

| # | what failed | cause |
|---|---|---|
| 1 | the workflow would not load at all | `shell: ${{ matrix.shell }}` — `shell` accepts no expression ([actions/runner#444](https://github.com/actions/runner/issues/444)). Neither pyyaml nor the SchemaStore schema objected; the rejection is GitHub's own |
| 2 | `"...node.exe C:\Program Files\nodejs\node.exe --version" did not report a version` | `Get-Command` returns **every** match and a hosted runner carries two `node.exe`; `$nodeBin` was an array |
| 3 | `could not fetch v9.9.9 ... git said: warning: --depth is ignored in local clones` | Windows PowerShell 5.1 promotes a native command's **stderr** to a terminating error under `$ErrorActionPreference = 'Stop'`, so git's *warning* aborted a clone that had succeeded |

A fourth failure was in the harness, not the script: the `exit 4` probe captured the installer's error message with `2>&1` in a `pwsh` step, and GitHub runs `pwsh` steps with `Stop` — the same hazard as #3, one layer out. Both probes moved to `cmd`, which has no error records.

## What the green run establishes

- the install runs under **Windows PowerShell 5.1** through `-File` with a positional version
- the shim carries its marker, is **CRLF**, and reports the version `package.json` declares
- a second install through the **documented `irm | iex` form** is recognised as an upgrade — which also answers whether a leading `param()` block survives `Invoke-Expression`: **it does**
- an occupied destination is refused with **exit 4**, foreign file unmodified
- an unknown tag is refused with **exit 2**, no checkout left behind
- `init` through the shim records `commitlore.bin` as the **`.mjs` bundle** and `commitlore.node` as an absolute interpreter; `context` works; `validate` refuses an invalid record and accepts a valid one

## What it deliberately does not establish

`doctor` reports three failures on Windows, and this job prints them without requiring them to pass:

```
commit-msg hook  fail  could not run the hook: spawnSync /bin/sh ENOENT
hook runtime     fail  could not run the hook: spawnSync /bin/sh ENOENT
PreToolUse       fail  configured hook executable "commitlore" is not resolvable from PATH
```

Two causes: the probe spawns `/bin/sh`, which Windows does not have, and the shim is `commitlore.cmd` while the configured entry names `commitlore`. Both belong to **T-1124 (#283)**, where they are now [recorded with the full output](https://github.com/MongLong0214/commitlore/issues/283#issuecomment-5141746698). Requiring a green `doctor` here would fail this ticket for another ticket's work, or smuggle in the Windows support claim this one is forbidden from making.

**Windows is not claimed as supported.** The installer works; the hook path is unestablished and named as such.

PR #320, opened 2026-07-31
Execute #71's containment on Windows, with the baseline that gives it meaning

MongLong0214:
For #283 (T-1124). **Draft, and stopping here by the ticket's own stop condition.** The measurement is complete; the assertions behind it do not pass, and nothing in this PR weakens them or claims Windows support.

## What was established, on `windows-latest`

| Question T-1124 had to answer | Executed result |
|---|---|
| Does Git for Windows run the hook with its bundled `sh`? | **Yes** — `/usr/bin/sh`, `MINGW64_NT-10.0-26100`. The `spawnSync /bin/sh ENOENT` in `doctor` is the **probe's** failure, not the hook's |
| Does Windows `init` record the same shape as macOS? | **Yes** — the `.mjs` bundle and an absolute interpreter |
| Is a recorded target outside the install root refused? | **Not established** — the baseline fails first, so this never ran |
| Is an unrecognised target refused? | **Not established** — same |
| Does a legitimate recorded target still work? | **No — the commit hangs and never returns** |
| Does PATH resolution find the `.cmd` shim? | **No** — `command -v commitlore: NOT found` |

The findings are recorded in full, with the runner output, in **#321**.

## The short version

```
commitlore.root = C:\Users\...\commitlore\v9.9.9        <- written by node, win32
dirname+pwd -P  = /c/Users/.../commitlore/v9.9.9/dist   <- read under MSYS sh, POSIX
containment case: NO MATCH -- the stub falls through
test -x node: yes
test -L bin: no
command -v commitlore: NOT found (the shim is commitlore.cmd)
real commit: HUNG -- killed at 90s, the hook never returned
```

Both of the containment guard's other conditions pass. Only the comparison fails, and it fails for every value including the installer's own bundle — so the guard neither contains nor permits, and control falls through to a walk that a probe on the same runner measured as `walk: NON-TERMINATING, stuck at [C:]`.

`doctor`'s TypeScript mirror reports **no problem** in that same state, because `node:path` is `path.win32` and handles the pair correctly. It is green exactly when the hook is dead.

## Why nothing is fixed here

`src/hooks/commit-msg.ts` is not in T-1124's **Owns**, and the ticket's stop condition is explicit: if the attacks fail on Windows, stop, do not weaken the job, do not mark Windows supported, and record the failure as its own issue. That is #321.

`docs/COMPATIBILITY.md` does not exist on `dev` (T-1122 / #271 has not merged), so this ticket's one-cell edit has no target. The row is left unflipped and this is the recorded reason — which is also the correct outcome on the evidence: Windows is reachable, not supported.

## What the job contains

Four steps appended to `install-ps1`, no `continue-on-error`, `timeout-minutes: 30`:

1. **Measurement, not assertion** — which `sh`, what `init` recorded, the containment `case` recomputed verbatim, where a fall-through lands, a scratch-repo probe reporting what a hook actually sees on this platform, and a bounded real commit.
2. **Baseline** — a valid record accepted *and* an invalid one refused. Runs before the attacks because an attack that "passes" when nothing resolves is not containment, it is a hook that refuses everything.
3. **Attack 1** — a recorded `commitlore.bin` outside the install root, with a positive control proving the witness fires before the attack relies on its absence.
4. **Attack 2** — a `commitlore.cmd` planted *inside* the root. Labelled in the source as a regression guard on a loosening a Windows fix would plausibly make, **not** as evidence that containment holds.

Every commit through the hook is bounded, and a timeout is reported as `HUNG` — distinct from a refusal — rather than aborting the step.

### Correction to an earlier commit message

`a57cbaf` explains the chained-hook probe's non-invocation as an argv0 with no `/` leaving `hook_dir` as `.`. The scratch-repo probe disproves that: `argv0 = .git/hooks/commit-msg` and `dir-from-argv0 = .git/hooks`. The probe's non-invocation has some other cause and is not diagnosed. It is also no longer load-bearing — the measurement no longer uses that slot.


MongLong0214:
## The steps in this PR pass once #321's fix is in front of them

Observed on a throwaway head (#326) that layers this job onto T-1127's fix — PR #327. `install-ps1`: **success**, steps 12-15 all green.

```
real commit: ACCEPTED (exit 0)
baseline ok: the hook runs, accepts a valid record and refuses an invalid one
positive control ok: the witness fires when the payload runs
--- commit exit 1 --- attack 1 refused: the program outside the install root never ran
positive control ok: the planted .cmd fires when it is run
--- commit exit 1 --- attack 2 refused: the unallowed-extension target never ran
```

**The two containment attacks have now executed for the first time.** Until this run the baseline failed first and they never ran at all. `commit exit 1` on both shows the hook ran and declined rather than the commit failing before it; the positive controls fire first, so an absent witness means not-executed rather than a witness this harness could never observe.

## Two lines in the measurement step are now stale, and they read as contradictions

Same log, same run:

```
containment case: NO MATCH -- the stub falls through
walk: NON-TERMINATING, stuck at [C:]
```

Both are wrong about the shipped code now, and both are this PR's to fix. They **restate** the pre-fix logic rather than deriving it: the measurement step hardcodes the old `case` and the old walk instead of reading them out of `commitMsgStub()`. So they will keep printing the original defect forever, next to a `real commit: ACCEPTED` that contradicts them.

That is the same class of mistake this job exists to catch — a check that restates what it claims to measure. T-1127's own test does it the other way and can be copied: it slices the walk out of `commitMsgStub()` and executes that text, so it cannot pass against a fix made only in a comment.

Suggested before this PR leaves draft: derive both diagnostics from the stub, then rebase onto `dev` once #327 merges and run the job at that exact head. Only that run — not this throwaway — is the evidence for Minimum GREEN.

Issue #321, opened 2026-07-31
Windows: the commit-msg hook hangs instead of returning, and #71's containment can never match there


MongLong0214:
Recorded from T-1124 (#283), which set out to establish #71's install-root containment for the wrapper path on Windows. The measurement answered a different question than the one it was written to ask, so it is recorded here rather than absorbed into that ticket.

**All of the below is from a `windows-latest` runner.** Nothing here is inferred from macOS behaviour.

## 1. The `commit-msg` hook never returns on Windows

A real `git commit` in a repository where `commitlore init` installed the hook does not fail and does not succeed. It hangs. In CI it was killed at a 90s bound; the job's cleanup then reported `Terminate orphan process: pid (7764) (sh)`, so the shell was still spinning after the commit was killed.

The cause is the last resolution step in the stub (`src/hooks/commit-msg.ts`):

```sh
dir=$PWD
while [ -n "$dir" ]; do
  if [ -x "$dir/node_modules/.bin/commitlore" ]; then exec ...; fi
  dir=${dir%/*}
done
```

`${dir%/*}` returns its input unchanged once no `/` remains. A probe hook installed in a scratch repository on the same runner reports what `$PWD` actually is inside a Git for Windows hook:

```
hook argv0    = .git/hooks/commit-msg
hook PWD      = C:/Users/RUNNER~1/AppData/Local/Temp/.../probe
hook pwd      = C:/Users/RUNNER~1/AppData/Local/Temp/.../probe
hook pwd -P   = /c/Users/runneradmin/AppData/Local/Temp/.../probe
hook uname    = MINGW64_NT-10.0-26100
walk: NON-TERMINATING, stuck at [C:]
```

The walk was executed with a counter on the runner. It does not terminate; it settles on `C:` and stays there. For a user this is worse than either accepting or refusing: the commit never comes back.

## 2. #71's install-root containment cannot match on Windows

The guard the walk is reached *through*:

```sh
recorded_dir=$(cd "$(dirname "$recorded")" 2>/dev/null && pwd -P) || recorded_dir=
case "$recorded_dir" in
  "$recorded_root"|"$recorded_root"/*) exec "$recorded_node" "$recorded" ... ;;
esac
```

Measured on the runner:

```
commitlore.bin  = C:\Users\...\commitlore\v9.9.9\dist\commitlore.mjs
commitlore.root = C:\Users\...\commitlore\v9.9.9
[...]

PR #326, opened 2026-07-31
TEMP (do not merge): T-1124's Windows containment steps against T-1127's fix

MongLong0214:
**Throwaway verification head. Not for merge — close after reading.**

The `install-ps1` containment steps belong to T-1124 (#320) and are not T-1127's to ship. This head is T-1127's fix (#321) with those steps layered on top, purely to observe what has never been observed: whether the baseline passes and whether the two containment attacks **execute at all** on `windows-latest`.

Until now the baseline failed first, so attack 1 and attack 2 have never run. What this run answers:

1. Does a real commit through the recorded install return, instead of hanging at the 90s bound?
2. Does the containment case **match** for the installer's own bundle, where it previously matched nothing?
3. With the baseline passing, do the two attacks run — and do they still **refuse**?

Point 3 is the one that matters most. A fix that made the legitimate path work by loosening the match would pass (1) and (2) and fail (3), and that trade is what T-1124's forbidden scope rules out.

The real gate is #320 rebased onto `dev` after #321 merges. This is the early read.


MongLong0214:
Read and done with. The result is recorded on #320: with T-1127's fix in front of them, `install-ps1` steps 12-15 all pass and the two containment attacks executed for the first time.

T-1127 merged as #327. The evidence for Minimum GREEN is #320 rebased onto `dev` and run at that exact head, not this throwaway.

Issue #484, opened 2026-08-08
0.7.0 shipped [directive] as unreachable: commander's [] default makes the nullish fallback to configured authors dead code


MongLong0214:
Released in 0.7.0, whose headline change was that `[directive]` became reachable. **It did not.**

## Reproduction against the released artefact

Fresh repository, one authored record, `commitlore.trustedAuthor` written the way `init` writes it:

```
$ node dist/commitlore.mjs inject --path pricing.ts
  [claim]      r-trust99  …  shared Redis cache | ops refuses another stateful dependency

$ node dist/commitlore.mjs inject --path pricing.ts --trusted-author owner@example.invalid
  [directive]  r-trust99  …  shared Redis cache | ops refuses another stateful dependency
```

The first form is what `CLAUDE_HOOK_COMMAND` runs. **Every record on every install still graded `[claim]`** — the exact condition [#415](https://github.com/MongLong0214/commitlore/issues/415) was opened about.

## Cause

`src/commands/inject.ts` declares the option with a default:

```ts
.option('--trusted-author <author>', '…', collect, [])
```

so `options.trustedAuthor` is `[]`, not `undefined`, when the flag is absent. The fix for #415 then wrote:

```ts
const trustedAuthors = options.trustedAuthor ?? configuredTrustedAuthors(cwd);
```

`[]` is not nullish. **The fallback never fired.** #415's defect, reintroduced one layer up by the fix for it.

## Why the tests passed

`test/trusted-authors.test.ts` drove `buildInjection` with options assembled by hand:

```ts
buildInjection({ path, cwd: repo, noIndex: true, trustedAuthors: configuredTrustedAuthors(repo) })
```

That asserts the grading layer works given correct options. It cannot see whether the CLI ever produces them, and the CLI did not.

**The file's own header had already written the warning**, one layer down:

> *"a unit test of `gradeRecord` would have passed throughout the entire period this bug existed"*

The same sentence applied to `buildInjection` and was not heard.

## Fix

Resolve on length rather than nullishness — an explicit flag is always non-empty, an absent flag is always empty, whichever shape commander hands over:

```ts
const flagged = options.trustedAuthor ?? [];
const trustedAuthors = flagged.length > 0 ? flagged : configuredTrustedAuthors(cwd);
```

Four new cases in `test/trusted-authors.test.ts` **spawn the built CLI** rather than importing into it. They fail against the 0.7.0 build.

## Also in the patch

[...]

Issue #472, opened 2026-08-08
doctor's runtime is unmeasured, and squash-conservation silently truncates its scan at 200 branches


MongLong0214:
doctor's own performance is an assertion with no measurement behind it, and one check quietly narrows what it looked at: `squash-conservation` slices its branch list at `MAX_SQUASH_CANDIDATE_BRANCHES = 200` (`.slice(0, 200)` in `squashCandidates`) and then reports over the subset without saying a subset existed. PRD §10.3 names that for what it is: "an `ok` over an unstated subset is the §8.3 rule violated in miniature."

## Why this exists

- PRD §10: "Unmeasured today; made measurable by §1.5, then fixed." The ADR's consequences section ties the move to precedent: per-check timing "is how the PRD's budget stops being an assertion (the same move as commit 1f8b4be made for the 100k-commit criterion)."
- ADR falsification 6 is the stake: a full check set measured slow enough that users skip running doctor reopens the profiles rejection — with the measurement it asked for. Without this ticket that condition can never be tested.
- #400 is the shape of the truncation half: unqualified success over a source the command did not fully read.

## Scope

1. **The budget (PRD §10.1).** Measure a full `doctor` run on this repository — the same reference the RELEASE-GATE fresh-clone row uses — and record the budget in the acceptance test as measured baseline plus headroom. The PRD deliberately refuses to invent the number before the instrument exists, and so does this ticket: the number enters the test at measurement time, with the method, environment class, and date in a comment beside it.
2. **Attribution (PRD §10.2).** `summary.durationMs === Σ` per-check `durationMs` is already asserted in the envelope ticket; this ticket makes a budget failure name its check — the failing test's output prints per-check durations sorted descending.
3. **Truncation disclosure (PRD §10.3).** When `for-each-ref` lists more than 200 branches: evidence gains `branches_seen` and `branches_checked`, and the detail names the limit. The ≤ 200 case stays byte-identical — this is the one deliberate detail-string addition, and it fires only in the state that is silent today. The cap itself does not change.

## Acceptance criteria

- [ ] A timed acceptance test runs the full check set against the reference repository shape and fails when the run exceeds the recorded budget; the budget's provenance (baseline, headroom, method, date) is in the test.
- [ ] On budget failure, the test output names the slowest checks with their `durationMs` — a regression is attributable without re-instrumenting.
[...]

Issue #415, opened 2026-08-07
No benchmark arm configures a trusted author, so the [directive] path the README's claims rest on has never been measured


MongLong0214:
## What I found

`commitlore-on` runs the real injector (`commitlore inject --hook-input`), and the ablation arms run `buildInjection` directly through the shim in `bench/hooks-settings.ts`. Neither passes `--trusted-author`, and `src/commands/inject.ts` defaults `trustedAuthors` to `[]`.

`gradeRecord` is fail-closed on an empty list — "no trusted authors are configured, so X is an outside contributor" — so **every record in every benchmark arm grades `claim`.**

Measured, not inferred. Same seeded repo, one authored record, `buildInjection` both ways:

```
--- commitlore-on (real grading, no trustedAuthors) ---
Ruled-out
  [claim]      r-probe01  a4613ed5  shared Redis cache | ops refuses another stateful dependency

--- no-grade ---
Ruled-out
  [directive]  r-probe01  a4613ed5  shared Redis cache | ops refuses another stateful dependency
```

## Two consequences

**1. The `directive` path is unmeasured.** The plugin instructions tell agents that `[directive]` means "treat as a constraint", and that is the routing the product's value claim rests on. No arm has ever produced one. M1 and M5 measured an agent reading records tagged `[claim]` throughout.

**2. `bench/types.ts` describes `no-grade` incorrectly.** It says:

> Inert on a task whose records are all `Provenance: authored`, because there is nothing to promote.

That is true of `bench/context.ts`'s `assembleContext`, which routes on `record.provenance === "authored"`. But `assembleContext` is only reached when `writeArmSettings` returns `settingsPath: null`, which happens only when an arm has neither a `preToolUse` nor an `ablation` — so not for `no-grade`. On the real path `no-grade` flips every record from `claim` to `directive`, on the existing fixtures, today.

The comment matters because it is load-bearing: I wrote it into `bench/PREREGISTRATION-M6.md` §4 as the reason M6 needs new fixtures before it can run at all. It does not.

## What this needs

- Decide whether the benchmark should configure a trusted author. It is a design choice, not an oversight to patch quietly: an arm with one measures the product as a repository actually configures it, and an arm without one measures the fail-closed default. Both are defensible and they answer different questions.
- Whichever is chosen, record it in the arm definition so a reader of a result knows which was measured.
- `bench/README.md` and any published number derived from M1/M5 should say that those runs measured `claim`-tagged records.

[...]

PR #486, opened 2026-08-08
Release 0.7.1 — the tier 0.7.0 announced, actually reachable

MongLong0214:
Promotion of `dev` into `main` for **v0.7.1**. Patch on the released 0.7.0.

## Why this exists

0.7.0 shipped with its headline change non-functional. The review verdict on that promotion was **FAIL** and named the defect; **the promotion had already been merged and tagged before the verdict arrived.** That sequencing failure is recorded in [#484](https://github.com/MongLong0214/commitlore/issues/484) and in the CHANGELOG, not only here.

## The defect, and the one the sweep found

`commander` declares `--trusted-author` with a default of `[]`, so `options.trustedAuthor ?? configuredTrustedAuthors(cwd)` never fell through. **Every record on every install still graded `[claim]`** — the condition #415 was opened about, reintroduced one layer up by its own fix.

The first commit's `Warn:` said it repaired one site rather than the class. Sweeping the class found `query.ts` immediately: `commitlore context` reported `claim` for a record the hook rendered `directive`, **while the comment at that line already asserted the two routes must answer alike.** A comment asserting a property is not the property.

## Verified against the released artefact

```
0.7.0   inject --path  → [claim]        context → [claim]
0.7.1   inject --path  → [directive]    context → [directive]
```

Six new cases **spawn the built CLI**. They fail against the 0.7.0 build, which is the only property that makes them worth having — the tests that passed drove `buildInjection` with hand-assembled options and never crossed the command line, which is the only path the hook uses.

## Also

- `package-lock.json` declared `0.1.0` while both manifests read `0.7.0`. Stale since the first release.
- The CHANGELOG corrects two miscounts in 0.7.0's own review evidence: **132 commits where the range held 137**, and **seven install checks where `RELEASE-GATE.md` §4 lists six**. Both were mine.

## Evidence at this exact head

Evidence is bound to the merged head. Any earlier head this PR carried has been replaced, and evidence from a head that no longer exists is not evidence.

```
merged head           ea66f50a16bfccfa6dd45abba526a37bd72a3c9a
merge commit          9a5a91a06bec50988f9cb6ea59f3e8003a08e773
range main..dev       19 commits, 32 files, +2665/-210
CI (push and PR)      success
promotion-range-lint  success
full suite            106 files, 2,308 cases pass, 1 skipped
manifests             0.7.1 in package.json, plugin.json, package-lock root and packages[""]
```

## Publication is gated, not sequenced by hand

Publication is gated by workflow dependencies, not by a procedure anyone has to
remember. `RELEASE-GATE.md` §4 is the `install-gate` job and `publish` depends on
it, so the checks that qualify a release run before the release exists (#493,
#495). Ancestry and exact-head CI are mandatory prerequisites too (#497), and
#499 binds every boundary to one canonical commit — anchored on the immutable
event SHA — so a tag cannot be moved between qualification and publication.

**No v0.7.1 tag or release exists.** The merge landed; publication is withheld
pending the #499 correction.

## 0.7.0 is not retracted

Its notes carry the defect at the top with a reproduction and a pointer here. Deleting a published tag breaks every clone that already fetched it.

## Stated costs

**Both routes now read git config on every invocation, including the hook on the hot path of every edit.** One local config read, but new work in a place measured in milliseconds — for #472's budget to bound.

The sweep covered `??` against an options field in `src/commands`, the shape that produced both defects. An option whose default is consumed some other way would not have shown up.


MongLong0214:
## Exact-head production gate — FAIL

**Head:** `ea66f50a16bfccfa6dd45abba526a37bd72a3c9a`  
**Base:** `1ec657184c3a5049cb3b8cb43e81e9e4edb684ec`

The exact-head focused tests (21/21), source and bench typechecks, deterministic build, full suite (106/106 files; 2,308 passed; 1 skipped), and required exact-head CI are green. Publication is still blocked by a release-reference TOCTOU defect.

At review time the repository had no tag ruleset and immutable releases were disabled. The workflow does not bind all four prerequisites and publication to one immutable commit:

- `version-consistency` uses the event SHA but compares only the tag name string.
- `release-target` fetches `github.ref` at that job's execution time and exports the commit it then observes.
- `exact-head-ci` checks that exported commit, while its own script checkout can re-fetch the tag later.
- `install-gate` re-fetches the tag and then clones again by tag name.
- `publish` consumes no commit output and runs `gh release create "$GITHUB_REF_NAME"`; a moved tag selects its new target, and a deleted tag can be recreated from the default branch because `--verify-tag` is absent.

A controlled ref-move regression recorded the original head from `check-release-target`, then moved `v0.7.1` to a different main descendant. The version and exact-head-CI gates still accepted the original SHA while the fresh tag clone resolved the new SHA. Therefore all required jobs can succeed without qualifying the object that publication ultimately selects; a move after all needs complete is sufficient.

The current regression tests assert only the four `needs` names and absence of a bypassing `if`; they do not assert common-SHA binding or moved/deleted-ref rejection.

The PR body is also stale: its “exact head” is `cb960d0e…` with 103 files / 2,271 cases, and its tag sequencing describes the superseded workflow rather than this head.

Required before re-review: protect release tags against update/delete before publication (immutable releases alone applies only after publication), propagate one canonical resolved commit through every checkout/clone/check/publish boundary, fail closed if the live tag differs or is absent, add moved/deleted-ref regressions, correct the PR body, and regenerate exact-head evidence on the corrected head.

No merge, tag, or release is authorized.

MongLong0214:
## Post-merge production gate — FAIL remains

PR #486 was merged at `9a5a91a06bec50988f9cb6ea59f3e8003a08e773` after the exact-head FAIL. The merge commit's second parent is the same failing head, `ea66f50a16bfccfa6dd45abba526a37bd72a3c9a`; no correction head replaced it.

The merge does not invalidate the technical blocker or turn prior green CI into release authorization. At post-merge readback:

- `v0.7.1` does not exist as a tag or release.
- there is no tag ruleset protecting `v*` from update/delete;
- immutable releases remain disabled;
- the common-SHA/publication-target TOCTOU described above remains present in the merged workflow;
- this PR body still reports stale exact-head and suite evidence.

Publication remains blocked. The bounded correction contract and adversarial acceptance cases are tracked in #499. No `v0.7.1` tag or release should be created until the correction has passed a new independent exact-head production gate.

PR #500, opened 2026-08-08
Publish the commit the gates qualified, not whatever the tag says now

### LINKED ISSUE #499: Bind release publication to one protected immutable tag target
## Failure

The release workflow can qualify one commit and publish a release whose live tag resolves to another commit.

At the failing review head `ea66f50a16bfccfa6dd45abba526a37bd72a3c9a`, the repository had no tag ruleset and immutable releases were disabled. The workflow mixed the tag-push event SHA with repeated tag-name resolution:

- `version-consistency` used the event SHA while comparing the tag name string.
- `release-target` resolved the live tag when that job ran.
- `exact-head-ci` checked the exported commit, but its script could fetch the tag again.
- `install-gate` fetched and cloned by tag name again.
- `publish` consumed no canonical commit output and created the release by tag name without `--verify-tag`.

A controlled ref-move regression demonstrated the split: the target and exact-head gates continued to accept the original SHA while a fresh clone of the same tag resolved a different main descendant. Deleting the tag also left a recreation path from the default branch.

PR #486 was subsequently merged as `9a5a91a06bec50988f9cb6ea59f3e8003a08e773` with the failing head as its second parent. No `v0.7.1` tag or release exists, so publication can still be withheld until this is corrected.

## Required correction

1. Resolve one canonical release commit once and propagate that exact SHA through all four prerequisite jobs and `publish`.
2. Make every checkout, script, and fresh-clone install gate consume the canonical SHA rather than re-resolving the tag name.
3. Immediately before publication, fail closed unless the live `refs/tags/<version>` exists and still resolves to the canonical SHA.
4. Create the release only from an existing verified tag (`--verify-tag`) and explicitly bind the release target to the canonical SHA.
5. Add an active repository tag ruleset for `v*` that prevents update and deletion before publication. Immutable releases may additionally protect published releases, but are not a substitute for pre-publication tag protection.
6. Correct PR #486's stale body evidence and sequencing description so it no longer names `cb960d0e…`, 103 files / 2,271 cases, or the superseded manual sequence.

## Tests and acceptance

- RED/GREEN regressions must move and delete the tag between prerequisite boundaries and immediately before publication; every case must refuse.
- A same-name tag resolving to a different main commit must refuse even when required CI passed on the original commit.
- A missing tag must refuse; publication must never create it implicitly.
- Tests must assert that all four prerequisite jobs and `publish` consume the same exported SHA, not merely that `publish.needs` lists four names.
- The active tag ruleset and immutable-release setting must be queried from GitHub and recorded as release evidence.
- Focused release tests, source/bench typechecks, deterministic build, full Node 24 suite, exact-head CI, and adversarial wrong-target/deleted-tag checks must all pass on the corrected head.

## Scope and sequencing

Release workflow, release-gate scripts/tests, repository release-tag protection, and correction of PR #486's body only. Do not push `v0.7.1`, create a release, or continue unrelated release work until an independent exact-head production gate records PASS on the corrected PR.

MongLong0214:
Closes #499. Base is the `main` merge commit `9a5a91a0`; exact head `46ab6569223fc7d5b0bf33f3a1b30bcdf297e9bb`.

**No `v0.7.1` tag or release exists.** Publication is held pending an independent exact-head gate PASS on this head.

## The failure

Four prerequisite jobs proved things about a commit, and **each one resolved the tag name separately to decide which commit that was.** A tag is a mutable ref, so two boundaries could get two answers — a ref move left the ancestry and CI gates accepting the original commit while a fresh clone of the same tag landed on another `main` descendant, every check green. The release could then ship a commit nothing had checked: the defect these gates were built to remove, rebuilt one level up by the gates themselves.

## The correction

The canonical commit is fixed once and consumed everywhere. **The anchor is the event SHA, not the live tag** — resolving the name in the first job would leave a window before that job starts in which a move redefines the whole release. The tag name survives only as the version label and as the key the final binding check looks up.

| boundary | consumes |
|---|---|
| `release-target` | `github.sha`, peeled to a commit |
| `version-consistency` | canonical SHA |
| `exact-head-ci` | canonical SHA |
| `install-gate` | clones `--no-checkout`, detaches onto the canonical SHA, asserts `HEAD` landed there |
| `publish` | canonical SHA, then refuses unless the live tag still resolves to it |

`scripts/check-tag-binding.mjs` queries **both** `refs/tags/<v>` and `refs/tags/<v>^{}`. `ls-remote` returns only the tag object for an annotated tag unless the peeled ref is named — confirmed against this repository's own `v0.7.0`, where one pattern gives `7f2aa4e2…` and only the peeled pattern gives the commit `1ec65718…`. A single pattern would have refused every annotated release by comparing its tag object.

The parser refuses ambiguity rather than resolving it: two sources, a repeated source flag, a remote alongside a seam, two rows for one ref, a peeled ref with no tag object, or a ref nobody asked for. Picking the convenient row would turn a guess into a verification.

## What is load-bearing, stated precisely

| interval | control |
|---|---|
| push → binding check | event-SHA propagation, then the binding check |
| binding check → `gh release create` | `v*` ruleset, `update` and `deletion` denied, no bypass actors |
| after the release exists | immutable releases |

**`--target` is not load-bearing** and nothing relies on it. `gh` documents it as the target used when the command creates a tag automatically; what the CLI does with it when the tag already exists is not asserted here either way. The proven property is narrow: it does not establish that the tag equals the canonical commit and cannot retarget an existing tag. It is passed as defence for one future mistake — if someone drops `--verify-tag`, the implicit creation that re-enables lands on the qualified commit rather than the default branch.

The binding check does **not** make the ruleset unnecessary. It catches drift that happened before it ran; the ruleset closes the window it cannot see.

## Evidence at this exact head

Clean worktree, clean `npm ci`, single run, Node v24.18.0:

```
full suite   107 files, 2,338 pass, 1 skipped
typecheck    clean
dist         rebuilds byte-identical
workflow     loads as YAML
```

17 of the first 20 new cases fail against the workflow as it stood. The same-name case drives the gates rather than their premises: ancestry and exact-head CI both pass for the original commit with six successful required check runs, **still pass after the tag moves**, and only the binding check sees that the tag no longer names it.

Repository controls, read back from the API:

```
immutable-releases   {"enabled": true, "enforced_by_owner": false}
ruleset 20587790     active · refs/tags/v* · [update, deletion] · bypass: []
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,14 +23,47 @@ permissions:
   checks: read
 
 jobs:
+  # The canonical commit is fixed ONCE, here, and every later boundary consumes
+  # that sha. Each boundary used to resolve the tag *name* again, which meant
+  # four gates could each qualify a different commit: a ref-move regression
+  # showed the ancestry and CI gates still accepting the original sha while a
+  # fresh clone of the same tag landed on another main descendant.
+  #
+  # The anchor is the event sha, not the live tag. `github.sha` is what the tag
+  # pointed at when the push happened and cannot be changed afterwards; the ref
+  # can. Resolving the name here would leave a window before this job starts in
+  # which a move redefines the whole release, so the tag name is only ever the
+  # version label and the key the final binding check looks up.
+  release-target:
+    runs-on: ubuntu-latest
+    outputs:
+      commit: ${{ steps.ancestry.outputs.sha }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.sha }}
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - id: ancestry
+        name: The pushed commit is already contained in main
+        run: node scripts/check-release-target.mjs "$GITHUB_SHA" main --github-output
+
   # Refuses the whole release before a single binary is built. A release
   # whose asset reports a version that disagrees with its own tag is not
   # fixable after the fact — the tag is immutable once fetched, the asset is
   # immutable once downloaded. See scripts/check-release-version.mjs.
+  #
+  # The checkout takes the canonical sha, not the tag: the version this job
+  # compares must come from the same bytes every other gate qualified.
   version-consistency:
+    needs: release-target
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.release-target.outputs.commit }}
       - uses: actions/setup-node@v4
         with:
           node-version: 22
@@ -39,27 +72,6 @@ jobs:
       - run: npm run build
       - name: Tag, package.json, and `commitlore --version` agree
         run: node scripts/check-release-version.mjs "$GITHUB_REF_NAME"
-
-  # A version-consistent tag can still name a commit that exists only on dev.
-  # `merge-base` can answer that only from complete history: a shallow checkout
-  # is not a smaller proof, it is an incomplete graph that the script refuses.
-  # The accepted commit is an output so the API gate below queries the commit an
-  # annotated tag resolves to, rather than assuming the event SHA is one.
-  release-target:
-    runs-on: ubuntu-latest
-    outputs:
-      commit: ${{ steps.ancestry.outputs.sha }}
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.ref }}
-          fetch-depth: 0
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 22
-      - id: ancestry
-        name: The pushed tag is already contained in main
-        run: node scripts/check-release-target.mjs "$GITHUB_REF_NAME" main --github-output
 
   # A tag push does not create a new CI result for the commit it names. The
   # relevant evidence is the existing set of check runs at that exact commit,
@@ -72,7 +84,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ needs.release-target.outputs.commit }}
           fetch-depth: 1
       - uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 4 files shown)


MongLong0214:
## Adversarial probe: glob ambiguity in the tag lookup — not a defect

`git ls-remote` treats its ref arguments as **patterns**, not literals, so `refs/tags/<version>` could in principle match more than the tag being verified. Probed rather than assumed:

A remote carrying both `v9.9.9` and `v9x9x9`, queried for `refs/tags/v9.9.9`:

```
f105c6c5  refs/tags/v9.9.9
```

One row. The `.` matches literally — git's ref patterns are fnmatch-style, where only `*`, `?` and `[` are wildcards.

And those three cannot appear in a tag name at all:

```
git check-ref-format refs/tags/v1*     rejected
git check-ref-format refs/tags/v1?0    rejected
git check-ref-format refs/tags/v1[0]   rejected
git check-ref-format refs/tags/v1.0    allowed
```

So a release tag can never carry a metacharacter, and the pattern is always effectively literal.

Closed twice over regardless: if a listing ever did return refs beyond `refs/tags/<v>` and `refs/tags/<v>^{}`, `check-tag-binding.mjs` exits 2 rather than reading a familiar row out of an unfamiliar payload. That is the `unexpected refs` refusal already covered by the suite.

**No change to the head** — `46ab6569223fc7d5b0bf33f3a1b30bcdf297e9bb` stands. Recording the probe because a checked assumption and an unchecked one look identical afterwards.

MongLong0214:
## Exact-head production gate: PASS

**Candidate:** `46ab6569223fc7d5b0bf33f3a1b30bcdf297e9bb`  
**Base:** `9a5a91a06bec50988f9cb6ea59f3e8003a08e773`

The #499 release-publication defect is closed on this exact head:

- one canonical commit is derived from the immutable tag-push event SHA and peeled to a commit once;
- version consistency, exact-head CI, the fresh-clone install gate, and publication all consume that exported SHA;
- publication re-reads the live tag immediately before release creation and refuses moved, deleted, missing, wrong-target, ambiguous, partial, or malformed bindings;
- annotated tags are compared by their peeled commit, not their tag object;
- `--verify-tag` prevents publication from recreating a missing reference;
- the active `refs/tags/v*` ruleset denies update and deletion with no bypass actors, and immutable releases are enabled;
- PR #486's stale exact-head evidence and sequencing description have been corrected.

Exact-head verification:

- focused release-tag regression suite: **30/30 PASS**;
- full suite: **107/107 files, 2,338 PASS, 1 SKIP**;
- source and benchmark typechecks: **PASS**;
- deterministic build and committed distribution comparison: **PASS, byte-identical**;
- conformance suite: **PASS**;
- exact-head GitHub CI on Node 22 and 24, install jobs, Git matrix, and PR lint: **PASS**.

Adversarial coverage includes pre-resolver ref movement, movement or deletion between boundaries, a same-name tag on another main commit while the original CI remains green, missing tags, lightweight and annotated tags, duplicate rows, unexpected refs, incomplete peeled listings, and malformed payloads. Each unsafe or ambiguous case fails closed.

**Verdict: PASS.** This exact head is approved for merge. After merge, release publication remains gated on successful CI at the merge commit; `v0.7.1` must target that verified commit, and the tag-triggered release workflow must complete successfully before the release is considered published.

PR #568, opened 2026-08-12
Close the eight blockers from the fourth review

MongLong0214:
All eight blockers from the fourth independent production-readiness review.

## 1 — the pre-edit hook stalled on any repository nobody had indexed

Installing the plugin arms the hook immediately; `commitlore init` is a separate per-repository step. In between, every edit paid for a full history scan that builds nothing, so the cost never amortised.

```
cold hook, no index   34.7s   per edit, indefinitely
warm hook              0.5s
```

The hook now scans under a wall-clock budget; commands a person ran and is waiting for keep none. A truncated answer is never silent — `unreadCommits` is a typed field for the same reason `notes` is, and the payload says how many commits went unread, warns that a decision shown as active may since have been withdrawn (supersession lives in commits too), and names `commitlore init`.

```
cold hook, no index    6.0s   and says what it skipped
warm hook              0.5s   unchanged, no notice
```

**The first version of this did nothing.** The deadline is checked between batches and `LOG_BATCH` is 1024 — more commits than most repositories have — so the whole scan was one batch and the check never ran twice. And the first test missed it: a budget that is *already spent* stops at the first check either way. The test now uses one that expires partway.

## 2, 5, 6 — a name was taken as evidence of the thing it named

| where | accepted | now |
|---|---|---|
| Codex marketplace | any marketplace called `commitlore` | its `marketplaceSource` must be ours; a foreign one is **refused, not replaced** |
| `.mcp.json` | the key's presence — `null`, `{}` all passed | an entry a host could actually launch |
| `install.ps1` | the name appearing in `codex mcp list` | `codex mcp get`, four-way, matching `install.sh` |

**A literal source comparison would have been worse than the bug.** This tool adds `MongLong0214/commitlore`; Codex reports `https://github.com/MongLong0214/commitlore.git`. String equality would call our own marketplace foreign and refuse on every correctly installed machine, so both sides reduce to host and path.

The `.mcp.json` check deliberately does **not** require our exact command — pointing the entry at a wrapper is a legitimate choice, and the callers respond to "not registered" by writing over it. It rejects only what nothing can launch.

## 3, 8 — results reported as established when they were not

`storeVerification` refuses any phase but `prepared`, and the refusal was discarded: a second verification returned draft B as accepted while staging used stored draft A. Now propagated — an unbound result is `empty` + `incomplete`. The catch-all's `incomplete: false` had the same shape and is now `true`, because an error is the definition of not knowing.

The MCP lifecycle log escaped into the working tree in a linked worktree: `join(cwd, absolutePath)` produced `<worktree>/Users/.../.git/worktrees/<name>/…`, in the exact case the function's comment promises to handle.

## 4, 7 — claims that outran their evidence

- The marketplace description said records "are ordinary git trailers, **so** they survive rebase, squash and rename". Being ordinary trailers is exactly why squash discards them (ADR-0004, `test/squash.test.ts`). Corrected on the marketplace and all four READMEs, naming `squash-preserve` as what does carry them.
- "Evidence-verified" → **"Quote-checked"**. Verification establishes that a quote *occurs*, not that it *supports* the trailer: `Limit: Redis is forbidden` citing `Redis is required` passes. `docs/capture.md` now states that gap with the example instead of concluding "so it is real".

## Verification

Every fix has a regression test, and each was confirmed load-bearing by restoring the defect and watching the new case fail. Typecheck and the separate bench typecheck clean; two builds byte-identical.


MongLong0214:
Amended: a `git add -A` in the last commit had swallowed `.commitlore-init-dist-LAZXXb/` — 258 files of a staging copy `commitlore init` makes while installing dist into a repository, left behind by a test run. Removed from the commit, and the pattern is now in `.gitignore` so it cannot recur. The branch is 44 files / 751 lines, which is the actual change.

Issue #633, opened 2026-08-13
prepare_capture fails with ENOENT on a dev-hash asset root that is never installed


MongLong0214:
`commitlore_prepare_capture` fails immediately and cannot be used at all:

```
commitlore: ENOENT: no such file or directory,
open '/Users/isaac/.local/share/commitlore/dev-3b70a1bebfb3/spec/SPEC.md'
```

The whole capture pipeline is unreachable, because `verify_capture` and `stage_capture` both need the nonce `prepare_capture` returns.

## What is on disk

```
$ ls /Users/isaac/.local/share/commitlore/
codex-plugin.json  v0.8.0  v0.8.1  v0.8.2

$ find /Users/isaac/.local/share/commitlore -maxdepth 3 -name SPEC.md
/Users/isaac/.local/share/commitlore/v0.8.2/spec/SPEC.md
/Users/isaac/.local/share/commitlore/v0.8.1/spec/SPEC.md
/Users/isaac/.local/share/commitlore/v0.8.0/spec/SPEC.md

$ ls /Users/isaac/.local/share/commitlore/dev-3b70a1bebfb3/
ls: No such file or directory
```

The payload directories are named by release version. The server resolves its asset root to `dev-3b70a1bebfb3`, which no installer step ever creates. So the lookup is not a missing file inside a valid directory — the directory itself does not exist.

## Reproduction

- macOS, `commitlore --version` → `0.8.2`, binary at `/Users/isaac/.local/bin/commitlore`
- MCP wiring is the default: `.mcp.json` → `{"command": "commitlore", "args": ["mcp"]}`
- Call `commitlore_prepare_capture` with any `transcript`. Repository policy is `{"mode": "auto", "unattended": true, "max_records_per_commit": 1, "require_verified_evidence": true}`, and the call passed `unattended: true`.
- Fails every time with the ENOENT above.

## What still works

`commitlore_query` answers normally, so the server starts, reads git history and serves records. Only the asset lookup is broken. That is consistent with the CLI and the MCP server disagreeing about which asset root to use rather than with a broken install.

## Why it matters

A repository can be fully wired — hooks installed, policy present, MCP connected — and still have no way to capture a record. The failure arrives at the moment of use, after the decision context has already been assembled, and it is not obviously distinguishable from "this commit has nothing worth recording", which is a normal and expected refusal. An operator following the documented flow will read a broken asset path as a rejection.

## What I could not determine

[...]

PR #636, opened 2026-08-13
Make doctor prove a registered MCP command is a healthy CommitLore server

### LINKED ISSUE #572: doctor reports a registered MCP command as working without establishing its identity
### Outcome
Doctor distinguishes "a command is present" from "a CommitLore MCP server is healthy".

### Acceptance
- initialize, serverInfo name/version, minimum tool set are probed
- dead/foreign/unverifiable command is not `ok`
- healthy custom wrapper is preserved

---

Fifth production-readiness review. Recorded for 0.8.0.

0.8.0 changed `.mcp.json` registration from "the key exists" to "the entry has a non-empty `command`", which is what stopped `null` and `{}` from counting. The remaining gap is identity: `{"command": "false"}`, or a wrapper an earlier install left behind, still counts as registered. `init` leaves it untouched and doctor reports `ok`, which claims a host can start unattended capture.

## The tension

Requiring our exact command would overwrite a wrapper somebody chose deliberately — the callers respond to "not registered" by writing our entry over it. Preserving a custom command is right. Reporting its identity as established is not.

## What to do

Keep preserving the entry; narrow what doctor claims. Report the command it found rather than a bare `ok`, or add a check that the registered command actually answers as a CommitLore MCP server.

`src/core/mcp-registration.ts`, `src/commands/doctor/checks/capture-unattended-initiator.ts`.

### LINKED ISSUE #635: commitlore_prepare_capture fails with ENOENT on a 0.6.0 spec path; the Claude plugin cache only has 0.8.0 and the codex cache has 0.8.2
The capture pipeline is unusable through MCP: `commitlore_prepare_capture` fails immediately, before any draft is written.

## What happened

```
commitlore_prepare_capture(transcript: "<~2.4 KB session note>")
→ commitlore: ENOENT: no such file or directory,
  open '/Users/isaac/.claude/plugins/cache/commitlore/commitlore/0.6.0/spec/SPEC.md'
```

There is no 0.6.0 anywhere on this machine. What exists:

```
$ ls -d ~/.claude/plugins/cache/commitlore/commitlore/*/
/Users/isaac/.claude/plugins/cache/commitlore/commitlore/0.8.0/

$ find ~/.claude/plugins/cache/commitlore -name SPEC.md
/Users/isaac/.claude/plugins/cache/commitlore/commitlore/0.8.0/spec/SPEC.md

$ find ~/.codex/plugins/cache/commitlore -name SPEC.md
/Users/isaac/.codex/plugins/cache/commitlore/commitlore/0.8.2/spec/SPEC.md

$ ls ~/.local/share/commitlore/
codex-plugin.json  v0.8.0  v0.8.1  v0.8.2

$ commitlore --version
0.8.2
```

So three different versions are live at once depending on which surface you ask:

| surface | version |
|---|---|
| `commitlore` on PATH | 0.8.2 |
| `commitlore.bin` in git config | v0.8.0 |
| Claude plugin cache | 0.8.0 |
| codex plugin cache | 0.8.2 |
| what `prepare_capture` looks for | **0.6.0** |

The 0.6.0 reference is the odd one — nothing else on the machine mentions it, so it looks like a
version string that is computed or persisted somewhere rather than read from the running install.

## Reproduction

1. Install 0.8.2 and wire the MCP server into a repository (this repo has `.commitlore-policy.json`
   and `commitlore.*` git config present).
2. Call `commitlore_prepare_capture` with any non-empty transcript.
3. Observe ENOENT on a `0.6.0/spec/SPEC.md` path.

The MCP server config in the plugin is:

```json
{ "mcpServers": { "commitlore": { "command": "node", "args": ["./dist/commitlore.mjs", "mcp"], "cwd": "." } } }
```

`cwd: "."` plus a relative `./dist/commitlore.mjs` means the resolved implementation depends entirely
on where the host launches it, which may be how the surfaces drifted apart.

## Impact

This blocks the documented capture flow end to end — prepare is the first step, so verify and stage
are unreachable. Commits with real decision context cannot be recorded through MCP at all; the only
路 left is writing trailers by hand, which the tool explicitly advises against.

## Possibly the same root cause as a separate report

I filed #631 for a related symptom: the same record reads `trust: "directive"` through the CLI and
`trust: "claim"` through MCP. If the MCP surface is resolving against a stale install, both would
follow from one cause. I am reporting them separately because I could not prove they share it.

## Environment

- CommitLore 0.8.2 on PATH; `commitlore.bin` git config pointing at v0.8.0
- macOS 15 (Darwin 25.3.0)
- Claude Code plugin cache at 0.8.0, codex plugin cache at 0.8.2

MongLong0214:
## Summary

Closes #572.

This fixes the live MCP identity check used by `doctor`. It verifies that the
registered command completes MCP initialization as CommitLore and advertises
the required read tools plus `commitlore_prepare_capture`,
`commitlore_verify_capture`, and `commitlore_stage_capture`.

A server that has only the two read tools is reported as a healthy
read-delivery server, but not as an unattended capture initiator. Launch,
identity, and incomplete tool sets remain distinct diagnostic outcomes.

## Boundary

This verifies MCP identity and advertised tool membership only. It does not
verify asset readiness or prove that `prepare_capture` will succeed: required
assets such as `SPEC.md` may still be missing. Asset readiness is the separate
F-002 finding tracked under #633 and broader runtime convergence #635; this PR
does not close #635.

## Test plan

- [x] `npx vitest run test/doctor.test.ts`
- [x] `npx vitest run test/init.test.ts`
- [x] Both doctor/init file orders
- [x] `npx vitest run test/installer-hosts.test.ts`
- [x] `npx tsc --noEmit`
- [x] Linux/amd64 `dist/` rebuild and canonical-artifact manifest refresh


MongLong0214:
## Verdict record for the install-ps1 flakiness

Kept here so a future recurrence starts from measurement rather than from scratch.

### Before the fix — 2 of 4 failed

| run | commit | install-ps1 |
|---|---|---|
| 31787895830 | `4aac882` | success |
| 31789044058 | `b917a42` | **fail** |
| 31789869220 | `5777066` | **fail** |
| (earlier) | — | mixed |

### After the fix — 4 consecutive successes

The fix is `47264ce` (probe budget 15s, sidecar bound derived from it, timeout mapped to *unverified* rather than *unhealthy*) together with `38f4b7d` (the fixture's launcher resolves node from PATH instead of a hard-coded path).

### What the diagnostics said

| | passing run | failing run |
|---|---|---|
| child saw stdin | **228ms after start** | never |
| pid file | present | absent |
| `durationMs` | **4478** of a 5000ms budget | ~5370 |

A *passing* probe consumed 89.6% of its budget. The chain is doctor → the sidecar's node → cmd.exe → the server's own node: four Windows process starts before a byte moves. Three independent drivers each got about five seconds and each fell silent, which reads as "not a timing problem" only if their budgets differed — they did not.

### Two hypotheses that were wrong, and how they died

- **"stdio does not traverse cmd.exe"** — refuted by driving the wrapper by hand, which answered.
- **"the launcher's hard-coded interpreter path is missing on this image"** — refuted by the passing run, which used that same launcher and recorded a child pid. The launcher change stays because resolving node from PATH matches what the rest of the job installs, but that is not what fixed this.

### Still open, deliberately

#657 — a launcher whose interpreter genuinely is absent is still reported as `initialize-timed-out` rather than as a command that could not run. Split out so it does not block this.

Issue #632, opened 2026-08-13
pre-push notes mirror reports ETIMEDOUT as a failure line when the push itself succeeded


MongLong0214:
Second observation from real use of v0.8.2 on `MongLong0214/agent-control-plane`. This one is minor and the behaviour is arguably correct — filing it for the wording and the retry story, not the fallback.

## What happened

```
$ git push -q origin main
commitlore: notes mirror (origin) failed: spawnSync git ETIMEDOUT; branch push continues
```

The branch push succeeded. Immediately afterwards:

```
$ commitlore sync
origin       nothing-to-do  no notes mirror on either side
```

So the mirror had nothing to publish in the first place, and the failure line was about a transient network timeout while checking.

## What is right about it

Not blocking the push is the correct call, and saying so on the same line (`branch push continues`) is genuinely useful — I did not have to guess whether my push landed.

## The friction

The line reads as though something was lost. In this case nothing was: there were no notes to mirror at all. An agent reading that output has to run `commitlore sync` to find out whether it needs to do anything, and a human is likely to either ignore it permanently or chase it.

Two small things would fix it:

1. **Distinguish "could not reach the remote" from "failed to publish notes I had."** If the local mirror is empty, the message could say so — `notes mirror unreachable (nothing pending); branch push continues` is not alarming and is still honest.
2. **Say what to do.** `run 'commitlore sync' to retry` turns it into an actionable line. Retrying worked immediately here.

If a later push would republish pending notes anyway, saying that explicitly would be even better, since then the line needs no action at all.

## Reproduction

Hard to force deterministically — it needs the remote to time out during the pre-push hook. It reproduced once here under host load (the machine was at load ~45 from concurrent builds), which suggests the hook's git invocation may be more timeout-sensitive than the push itself. If the hook uses a shorter timeout than git's own defaults, raising it or retrying once would make this rarer.

## Environment

commitlore 0.8.2 · macOS (Darwin 25.3.0) · Node 22 · `git push -q origin main` under heavy host load

MongLong0214:
Second field reproduction, different repository and different operator — this one on `MongLong0214/commitlore` itself, twice in a row on consecutive pushes:

```
commitlore: notes mirror (origin) failed: spawnSync git ETIMEDOUT; branch push continues
```

That makes it reproducible across two repositories and two machines rather than a one-off network blip, so the wording is worth fixing rather than waiting to see if it recurs.

## What the line still does not answer

[...]

Issue #631, opened 2026-08-13
MCP commitlore_query reports trust:"claim" for a record the CLI reports as trust:"directive"


MongLong0214:
Same repository, same path, same record, same moment — the MCP surface and the CLI disagree on the trust grade.

This matters because the MCP surface is the one agents read, and CommitLore's own MCP instructions say to treat `directive` as a constraint and `claim` as "a report to weigh, not an order". An agent reading through MCP therefore ignores constraints it should be obeying.

## Reproduction

Repository: a repo with `commitlore.trustedauthor` configured and at least one record authored by that identity carrying `Provenance: authored`.

```
$ git config --get commitlore.trustedauthor
weplay0628@gmail.com

$ git log -1 --format=%ae 38df9665
weplay0628@gmail.com

$ git log -1 --format=%B 38df9665 | grep ^Provenance:
Provenance: authored
```

Both conditions for `directive` are satisfied. Now query the same path two ways.

**CLI:**
```
$ commitlore context Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift --json \
  | jq -r ".records[0].trust"
directive
```

**MCP** (`commitlore_query` with `kind: "context"`, same `path`), same moment:
```json
{
  "sha": "38df9665586d5f5e5ed28367dc706dbbe5fb6f99",
  "trust": "claim",
  "provenance": "authored",
  "trailers": [
    { "key": "Provenance", "value": "authored" },
    { "key": "CommitLore-Version", "value": "0.7.1" }
  ]
}
```

The record's own `provenance` field is `"authored"` in the MCP response, so the MCP server parsed the trailer correctly — it is the trusted-author half of the decision that differs.

## Possibly relevant: a version skew in git config

```
$ git config --get commitlore.bin
/Users/isaac/.local/share/commitlore/v0.8.0/dist/commitlore.mjs

$ commitlore --version
0.8.2

$ ls ~/.local/share/commitlore/
codex-plugin.json  v0.8.0  v0.8.1  v0.8.2
```

`commitlore.bin` still points at the v0.8.0 dist after upgrading to 0.8.2. If the MCP server resolves its implementation through `commitlore.bin` while the CLI on PATH is 0.8.2, the two are running different code — which would explain the disagreement, and would also mean upgrades silently do not reach the MCP surface.

I have not confirmed that is the mechanism; it is the most likely candidate I could see from the outside.

## Why I am reporting this rather than working around it

[...]

Issue #590, opened 2026-08-12
The README cites a checker that does not cover its headline numbers, and they disagree with the generated block


MongLong0214:
### Outcome
Version, Node floor, headline metrics, provenance vocabulary, and compatibility facts come from one generated/checkable source shared by English and Korean docs.

### Acceptance
- conflicting denominators cannot pass CI
- every product-written provenance value appears in the docs
- checker states exactly which facts it owns

---

## What the README claims

It prints **2.8% (16/580)** against **18.8% (109/579)**, then says a statistic retyped into prose drifts and that `scripts/check-readme-numbers.mjs` gates against exactly that (README:257-273).

## What the checker covers

Only the `<!-- BENCH:BEGIN -->`…`<!-- BENCH:END -->` block, plus four stray shapes outside it: `p =`, `pp`, "odds ratio", "Fisher exact" (`scripts/check-readme-numbers.mjs:105-110`). Rates, `N/N` fractions, milliseconds and run counts are invisible to it.

So the paragraph explaining why the numbers can be trusted is describing a guarantee that does not extend to the numbers next to it.

## And they have already drifted

Same shards, three denominators, in one file:

| source | on | off |
|---|---|---|
| prose (README:257-264, matches `bench/VERDICT-M5.md:13-14`) | 16/**580** = 2.8% | **109**/579 = 18.8% |
| generated block (README:547-552) | 16/**585** = 0.027 | **110**/584 = 0.188 |
| generated Fisher row | 16/**569** | 110/**474** |

The off-arm count is 109 in prose and 110 in the block. The headline figure is consistent with a document that is not the one the gate regenerates.

Also outside the block and unchecked: 81.7% / 42.0% / 92.3%, 496 ms / 185.85 ms / 102.40 ms, 44.8% / 22.0%, 10,002, 7.4 seconds, 100,000 commits. `v0.8.1` and `Node.js 22.13` happen to match `package.json` today; nothing keeps them matching.

## Related: the README's Provenance vocabulary omits what the product writes

README:502 lists `authored | inherited <sha> | reconstructed`. The spec and the schema also accept **`drafted`** and **`unknown`** — and unattended capture stamps `Provenance: drafted` on every record it produces (`src/core/capture-verify.ts:390-396`). `docs/protocol.md` calls that README table the working summary of SPEC §3.

[...]

Issue #638, opened 2026-08-13
amend that fixes a malformed trailer is refused as a duplicate Record-Id


MongLong0214:
The commit-msg hook correctly warns that a wrapped `Limit:` line is not a trailer, then refuses the `git commit --amend` that fixes it. The only way out is `--no-verify`.

## Sequence

Commit with a trailer block where two `Limit:` values are wrapped across two lines each, so they sit in their own paragraph above the real trailer block:

```
Limit: Fixtures intentionally copy the live authority tree; replacing that
structure with a wholly synthetic fixture is outside D0-013 exact ownership.
Limit: cleanupFixture intentionally suppresses cleanup errors so teardown
cannot obscure the test result; the known teardown-race tradeoff remains.

Warn: render witnesses depend on the live catalog and authority-tree fixture shape
Blast: local
...
Record-Id: r-d0013executionviewshardening
```

The hook lets the commit through with an accurate warning:

```
shape ok · references ok
commitlore: line 15 looks like a Limit trailer, but git did not parse it;
            the trailer block needs a blank line before it
commitlore: line 17 looks like a Limit trailer, but git did not parse it;
            the trailer block needs a blank line before it
```

`git log --format='%(trailers:only=true,keyonly=true)' -1` confirms it: `Warn Blast Undo Certainty Record-Id Follows Provenance X-Ticket`. The two `Limit` values are absent from the record.

Amend the message to put both `Limit:` values on single lines inside the final block, keeping the same `Record-Id`:

```
$ git commit --amend -F fixed-message.txt
shape ok · references failed
21: duplicate-id Record-Id — got "r-d0013executionviewshardening",
    want "exactly one record per Record-Id"
commitlore: 1 violation (SPEC §6) — the message was not modified
```

## Why this is wrong

`--amend` replaces the commit. After it, the old commit is unreachable from the branch and there is exactly one commit carrying that `Record-Id`. The duplicate the check reports does not exist in the resulting history — it exists only because the pre-amend commit was already recorded and the check does not know it is about to be replaced.

The two rules combine into a trap. The first rule tells you the message is malformed. The second rule blocks the correction, unless you either change the `Record-Id` — which is semantically wrong, since it is the same record — or bypass the hook. I used `--no-verify`, which is the outcome the hook exists to prevent.

## Reproduction

- macOS, `commitlore --version` → `0.8.2`
[...]

PR #643, opened 2026-08-14
docs: correct four factual defects in the README, and fix the hero SVG

MongLong0214:
## What this PR became, and why it is smaller than it started

It began as the A–Z review's README restructure. **The restructure cannot be done in this PR, and the reason is a contract, not a preference.**

`test/readme-order.test.ts:141,174` binds section order across all four language files:

```js
const FILES = ['README.md', 'README.ko.md', 'README.ja.md', 'README.zh-CN.md']
it('all four files have consistent relative order', ...)
```

`T-1016` requires the same demo asset in all four; `T-1015` requires the exposure table, the delivery/capture naming for every host class, and the clone-carried vs notes-backed distinction in every file. Restructuring English alone failed 18 tests across 5 files. Restructuring all four would take over the translated READMEs, which belong to #590.

So the restructure is inherently a coordinated four-file change and belongs with #590, not ahead of it. **What survives here is every factual correction the review found** — those needed no structural change, and they are the part that was actually wrong.

## The four corrections

**1 — the headline denominators contradicted the generated block.** The prose claimed `16/580` and `109/579`; the checker-owned block in the same file reports `585` and `584` runs per arm, with `110` in the off arm. Two different denominators for one study, in one file.

```
prose (unchecked)      2.8% (16/580)  ·  18.8% (109/579)
generated (checked)    585 and 584 runs per arm; commitlore-off 110
corrected              2.7% (16/585)  ·  18.8% (110/584)
```

The prose drifted because nothing owned it — which is exactly the gap #590 is widening.

**2 — the Capture column's "Yes" read as deterministic autocapture.** What is installed is the prepare → verify → stage path. That every eligible commit reaches a terminal assessment on its own is a stronger property, certified on no host. The caption now separates available from guaranteed.

**3 — the `Provenance:` vocabulary was missing two of its five values.** [SPEC §3](spec/SPEC.md) carries `authored | drafted | inherited | reconstructed | unknown`; the table listed three. `drafted` is what unattended capture writes, so the omitted value was the one a reader is likeliest to meet.

**4 — "Every route uses the same trust grading" stated an intention as an observation.** A record has been observed grading `directive` through the CLI and `claim` through MCP with more than one runtime installed ([#631](https://github.com/MongLong0214/commitlore/issues/631), [#635](https://github.com/MongLong0214/commitlore/issues/635)). Users meet the result, not the design.

## The hero SVG

grok's adversarial review found six defects in a version I had already accepted. Two I had missed: the agent appeared nowhere in the image, so "delivered to the next agent" was asserted rather than shown; and 19px text rendered at 5.9px at mobile width. Now: 22px floor (24px for required elements), `aria-labelledby` + `aria-describedby`, no strikethrough, and the superseded record is a genuinely different record rather than the same one crossed out.

## Verification

```
bash spec/verify.sh                    OK: 32 fixtures + README example sync + vocab table
node scripts/check-readme-numbers.mjs  block matches bench/report.ts
vitest readme + positioning + numbers + order + compatibility + demo   107 passed
```

## Filed separately

- **#645** — SPEC §7 still describes signature mode as `G` status only, while #642 also requires the `%GF` fingerprint in `commitlore.trustedSigner`. Every other surface picked that up; the normative text did not.
- **#632** — `notes mirror (origin) failed: spawnSync git ETIMEDOUT` reproduced on every push here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,12 +200,18 @@ separate layers:
 | Gemini CLI, Cursor, Windsurf, opencode | **Yes — the MCP server is wired by `install.sh`.** | **Procedure, not automatic.** The server states the prepare → verify → stage procedure in its `instructions` on every connection. The host may or may not act on it. |
 | Any other `AGENTS.md`-convention host | **Procedure, not automatic.** `commitlore init --agents-md` writes it into the repository. | **Procedure, not automatic.** Same file, same caveat. |
 
-“Yes” means the layer is installed, not that every commit gains a record.
-Most commits should carry none. The first three rows install a skill that
-drives capture; the fourth receives the same procedure over MCP, which is what
-a host that loads no skills has to work from — verified with the plugin
-disabled, and it captured. Whether a given host surfaces those instructions to
-its model is the host's choice, and nothing here detects it. A host still has to start capture, and the candidate must pass
+“Yes” in the Capture column means the workflow is installed and available — the
+prepare → verify → stage path exists for that host. It does not mean every
+eligible commit reaches a terminal assessment on its own. **That stronger
+property, deterministic autocapture, is not certified on any host yet**, so
+treat capture as available rather than guaranteed. Most commits should carry no
+record in any case.
+
+The first three rows install a skill that drives capture; the fourth receives
+the same procedure over MCP, which is what a host that loads no skills has to
+work from — verified with the plugin disabled, and it captured. Whether a given
+host surfaces those instructions to its model is the host's choice, and nothing
+here detects it. A host still has to start capture, and the candidate must pass
 verification before the commit hook attaches it. The commit-msg hook validates
 a record when present; it never invents one.
 
@@ -270,12 +276,12 @@ rather than in a review comment after.
 
 **Whether it acts on that is now measured.** Across 1,160 registered runs, an
 agent handed the repository's active records re-proposed a ruled-out approach in
-**2.8%** of them (16/580). Without them: **18.8%** (109/579).
+**2.7%** of them (16/585). Without them: **18.8%** (110/584).
 
 | arm | re-proposed a ruled-out approach |
 |---|---:|
-| the agent alone | **18.8%** (109/579) |
-| **with CommitLore** | **2.8%** (16/580) |
+| the agent alone | **18.8%** (110/584) |
+| **with CommitLore** | **2.7%** (16/585) |
 
 **The threshold was registered before the run**, and the preregistration
 predicted a *smaller* effect than it got — that prediction, with its stated
@@ -513,7 +519,7 @@ CommitLore-Version: 2.0.0
 | `Follows:` / `Supersedes:` | Decision-chain and lifecycle links |
 | `Expires:` | Date or condition that ends a limit |
 | `Evidence:` | Path, anchor, or URL supporting a claim |
-| `Provenance:` | `authored` \| `inherited <sha>` \| `reconstructed` |
+| `Provenance:` | `authored` \| `drafted` \| `inherited <sha>` \| `reconstructed` \| `unknown` |
 | `CommitLore-Version:` / `X-*:` | Protocol identity and extensions |
 
 Read a path's history with `commitlore context <path>`. Smaller examples, and how to read records with plain Git instead, are in [docs/protocol.md](docs/protocol.md); the normative definitions are in [SPEC §3](spec/SPEC.md).
@@ -521,7 +527,7 @@ Read a path's history with `commitlore context <path>`. Smaller examples, and ho
 ## What the repository proves
 
 - Decision history survives rebase, remote transfer, and path renames in the tested Git workflows. Squash-merge discards the trailer block, as any ordinary trailer would be: `commitlore squash-preserve` or its GitHub Action carries the records across, and the tested workflows cover that route.
-- Every route uses the same trust grading, so untrusted text is information rather than an instruction.
+- Routes share one grading core, so untrusted text is information rather than an instruction — but a record has been observed grading differently through the CLI and through MCP when more than one CommitLore runtime was installed ([#631](https://github.com/MongLong0214/commitlore/issues/631), [#635](https://github.com/MongLong0214/commitlore/issues/635)).
 - Injection-like text in free-form trailers is withheld from model-readable routes.
 - A readable repository with no records is distinct from incomplete history or an unfetched notes mirror.
 
--- a/assets/readme/hero.svg
+++ b/assets/readme/hero.svg
@@ -1,26 +1,17 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="1200" height="400" viewBox="0 0 1200 400" role="img" aria-labelledby="title desc">
-  <title id="title">CommitLore — Your agents inherit the code. Make them inherit the judgment.</title>
-  <desc id="desc">A fresh coding agent receives the active decision for a path, carrying a ruled-out alternative and its reason, while a superseded record is withheld because a later decision reversed it.</desc>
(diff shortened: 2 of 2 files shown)


MongLong0214:
## grok adversarial review → six fixes (ffc8f1e)

The SVG was built by grok against the handoff spec, then reviewed adversarially by a fresh grok pass against the same acceptance criteria. It failed two of them. Both are invisible at desktop width, which is why the review was worth running.

| # | Severity | Defect | Fix |
|---|---|---|---|
| 1 | High | **The recipient was missing (AC-01).** File, ACTIVE, SUPERSEDED and "not delivered" were all present, but nothing named the agent, so "it gives an *agent* the current decisions" was not readable from the image. `<title>` also said CommitLore "attaches" decisions to a path — the wrong mechanism. | `delivered to the next agent`; title and desc rewritten to delivery-to-an-agent |
| 2 | High | **Unreadable at the specified size (AC-03).** 19px labels render at ~5.9 CSS px at 375px, and GitHub's mobile column is nearer 330px. Secondary ink was 4.7:1. | floor raised to 22px, required elements 24px, secondary ink darkened, viewBox 320→340 to buy room |
| 3 | Medium | **Same idea twice.** Active `Ruled-out — Reuse checkout pricing for admin quotes` next to superseded `Reuse calculatePrice for admin quotes` read as one decision printed twice, i.e. "it still shows the old rule". | superseded is now a distinct earlier record: `Earlier Limit — calculatePrice also served admin quotes` |
| 4 | Medium | **Strikethrough read as prohibition.** A line through the superseded action, beside an active Ruled-out of the same action, claims closer to blocking than the product does. | strikethrough removed; superseded carried by position below the divider, a quieter label and lower contrast — more than colour alone |
| 5 | Low | Strikethrough overshot its text by ~80px | moot with 4 |
| 6 | Low | `aria-labelledby="title desc"` concatenated both into one accessible name | `aria-labelledby="title"` + `aria-describedby="desc"` |

Passing unchanged: one art direction, no terminal chrome, no gradient, 5 colours, no marketing headline inside the drawing, fully static, no external font/image/JS/`foreignObject`, no forbidden wording, 35 visible words, every run inside the 1200 viewBox with margin.

Defects 3 and 4 are the ones worth noting beyond this PR: both were the drawing claiming more than the product contract allows, which is the same class of error `P0-README-02` fixed in the prose.

MongLong0214:
## Conformance restored, and the vocabulary drift closed with it (f70127a)

The first push failed `git-matrix` on macOS and ubuntu. That was the contract working, not a flake.

`spec/verify.sh` reads the README's complete-record code block **as a conformance fixture** and compares the vocabulary table against SPEC §3. Moving both to `docs/protocol.md` — which the review recommended — silently deletes those two checks rather than relocating them. Both are back, placed at the end so the record grammar no longer competes with the product entry.

Restoring the table also closed **P0-README-04**. It listed `Provenance: authored | inherited <sha> | reconstructed` while [SPEC §3 line 124](spec/SPEC.md) carries `drafted` and `unknown` as well — and unattended capture writes `drafted`, so the missing value was the one a reader is most likely to meet.

## Scope confirmed with the CommitLore CTO

**The three translated READMEs are #590's, not this PR's.** Hand-syncing them would route around the very drift #590 exists to fix. The boundary agreed:

| #590 generates into all four | This PR owns |
|---|---|
| Provenance vocabulary values | table headers and surrounding prose |
| bench figures, denominators, rates | the sentences that frame them |
| Node floor, version | section structure and placement |
| capture certification state values | the wording that explains the state |

The vocabulary correction is already in the #590 branch for all four languages. Merge order **this PR → #590 rebase** closes it without duplicate work, at the cost of a short mismatch on `main`.

**Neither the generated block nor the vocabulary table is leaving the README.** #590's goal is widening the set of facts the checker owns, not relocating them — the block stays where this PR placed it and #590 fills it, reusing the existing marker names and format. No follow-up PR to strip markers is needed.

## Verification

```
bash spec/verify.sh                    OK: 32 fixtures + README example sync + vocab table
node scripts/check-readme-numbers.mjs  block matches bench/report.ts (54 lines, 4573 bytes)
in-page anchors                        6/6 resolve
```

606 → 352 lines.

## Filed separately

`commitlore: notes mirror (origin) failed: spawnSync git ETIMEDOUT; branch push continues` reproduced on both pushes here and has been folded into **#632** with the reproduction. Fail-open is by design; the message not saying whether a note is pending, whether the next push retries, or what to run is the open product decision.

MongLong0214:
## CI will not go green — two failures on `check (24)` and `check (22.23.2)`

Flagging now rather than letting the background watch wait for a green that is not coming.

### 1. Compatibility pointer count

```
× T-1122 the compatibility statement exists and is the authoritative one
  > README.md points at it exactly once
  → README.md must carry exactly one pointer: expected 6 to be 2
```

The restructure introduced more references to the compatibility document than the test allows. That test exists to keep one authoritative pointer rather than several drifting ones, so the fix is to reduce the references, not to relax the count — same reasoning as the conformance fixture you already restored.

### 2. Four commits carry a record with no `Record-Id`

```
× dogfooding: this repository obeys its own protocol > every in-scope record carries a Record-Id
  a0a7b83 docs: make the README a product entry, and say what capture does not do yet
  c5437ba docs: fix six defects grok's own review found in the hero SVG
  6fe4215 docs: keep the conformance fixture in the README, and complete the vocabulary
  ff33333 docs(readme): 등급 일관성 문장을 실측에 맞춰 정정
```

Each carries CommitLore trailers but no `Record-Id:`, and the protocol requires one per record.

**This is the amend trap in #638.** Adding the trailer means amending, and the hook then refuses the amend as a duplicate `Record-Id` against the commit it replaces — the escape today is `--no-verify`, which is exactly what #638 says teaches people to bypass the gate. If you hit that, use `--no-verify` for the amend and say so in the PR; the gate still validates the resulting message on the next real commit. I would rather record that we hit it than have it quietly shape how the history looks.

Reproduce both locally with:

```
npx vitest run test/readme.test.ts test/dogfood.test.ts
```

MongLong0214:
## New failure after the scope reduction — the record you added collides with an unrelated proposal

```
× guard against this repository > does not flag: document the exit codes in the README
  → expected [ '0.625 README를 제품 진입점으로 전면 재구성' ] to deeply equal []
```

Guard now treats **"document the exit codes in the README"** as an attempt to revive a ruled-out approach, at 0.625.

## Why

The alternative you recorded is broad enough to swallow any README proposal:

```
Ruled-out: README를 제품 진입점으로 전면 재구성 | T-1015/T-1016이 4개 파일 일관성을 강제하는데 번역은 #590 소유라 이 PR에서 손댈 수 없다
```

The *reason* is precise — four-file consistency, translations owned by #590. The *alternative* is not: as written it reads as "restructuring the README", so anything about the README scores against it.

That test is a precision floor for guard. Guard is an experimental advisory with measured precision well below a gate (ADR-0020), and this case exists so it does not drift further. Relaxing it to make the PR pass would spend the floor to keep a record's wording.

## The fix belongs in the record, not the test

Name what was actually rejected. It was not "restructuring the README" in general — it was restructuring **only the English file** while section order is a four-file contract. Something closer to:

```
Ruled-out: restructuring README.md alone, leaving the three translations on the old section order | T-1015 binds relative section order across all four files, so an English-only restructure cannot pass, and the translations belong to #590
```

That records the same decision, is more useful to the next person — it says what *would* be acceptable, a four-file change — and stops scoring against proposals that merely mention the README.

Worth noting the record did its job here in an unexpected way: it is the first time in this session that a recorded decision has actively pushed back on a later proposal. The pushback was wrong, and the reason it was wrong is that the alternative was written as a category rather than as the thing that was tried.

Reproduce with `npx vitest run test/guard.test.ts`.

Issue #430, opened 2026-08-07
A commit that carries a record can never be amended: the duplicate-id check counts the commit being replaced


MongLong0214:
## What happens

Commit anything carrying a `Record-Id`, then amend it — for any reason at all:

```
$ git commit --amend --no-edit
shape ok · references failed
4: duplicate-id Record-Id — got "r-amend01", want "exactly one record per Record-Id"
commitlore: 1 violation (SPEC §6) — the message was not modified

git exit code: 1
HEAD changed: NO
```

`--no-edit` is the important line. The record is not being changed, re-declared, or touched. The amend is rejected anyway.

Adding a forgotten file is rejected the same way:

```
$ echo b > g.ts && git add g.ts && git commit --amend --no-edit
exit 1
g.ts in HEAD: 0
```

**A commit that carries a record cannot be amended. Ever.** Not to fix a typo in the subject, not to stage a file left out, not with `--no-edit`.

## Why

At `commit-msg` time during an amend, HEAD is still the commit being replaced. The reference check walks history, finds `r-amend01` on HEAD, sees the same id in the incoming message, and reports a duplicate — correctly, by its own rule. Nothing tells it that the commit it matched is the one about to cease existing.

## Why it matters more than it looks

Amend is one of the most common operations in git, and it is the standard fix for the two most common mistakes a person makes at commit time.

It gets worse with unattended capture (#426): once capture stages a record without asking, **every** recorded commit becomes unamendable, and the user did not choose to put a record there.

## Scope

- Rebase and merge are **fine**. Verified: three recorded commits rebased onto root and merged with a recorded merge commit; all records survive and `context` reports them correctly. This is specific to amend, where the old and new commit coexist at hook time.
- The identity rule itself is right. SPEC §3.2 and the `Record-Id` uniqueness check exist for a reason, and #409's grading work depends on identity being trustworthy. The defect is the scope of the walk, not the rule.

## What a fix has to avoid

Simply excluding HEAD from the duplicate walk would let a genuine duplicate through whenever the colliding record happens to sit on HEAD. Whatever the fix is, it needs to distinguish "HEAD is being replaced by this commit" from "HEAD is a different commit that already used this id" — and `commit-msg` receives no argument saying which. `prepare-commit-msg` does get a source argument; whether that is a usable signal, and whether it survives `git commit --amend` in every form, is the first thing to establish.

MongLong0214:
Design survey, so whoever fixes this does not repeat it.

[...]

PR #663, opened 2026-08-14
Stop counting HEAD's record when validating a commit that does not exist yet

MongLong0214:
**Do not merge without the owner's decision.** This narrows a protocol rule in SPEC §6, which is not a ticket's call. The implementation is here so the decision is "merge this or not" rather than "design this".

Addresses #638.

## The defect

An amend that repairs a malformed trailer is refused as a duplicate. Correcting a trailer is the one edit that changes the payload, and one `Record-Id` with two payloads is a collision — so the rule fires precisely on the repair. The way out was `--no-verify`, which is the outcome the hook exists to prevent.

## Why the obvious fix does not exist

The rule that should apply is *ignore a declaration that will not remain in history*; during an amend, HEAD is exactly that. **A commit-msg hook cannot observe whether it is amending.** No environment variable distinguishes it, the message file is `COMMIT_EDITMSG` either way, and comparing against HEAD's message fails the moment the author edits it — which is this case.

## The three options, and what killed two of them

**A — do not count HEAD.** Implemented here.

**B — count HEAD only when payloads differ and the new message is not a superset.** A heuristic about editing would decide whether a record directs an agent. **No real superset case exists**; neither I nor the reviewer could produce one. Recording that, so nobody revives it as "a heuristic would do".

**C — have `prepare-commit-msg` record the amend for `commit-msg` to read.** The only option that observes rather than infers, and it was rejected on a measurement rather than on taste. Hook stubs, three paths:

```
1  prepare src=message      ← git commit -m
2  commit-msg ran
3  prepare src=message      ← git commit --no-verify -m      (prepare only)
4  prepare src=             ← GIT_EDITOR=true git commit (empty message)
5  commit-msg ran
```

**Line 3 ends C.** `--no-verify` runs the hook that writes the marker and skips the hook that consumes it, so the marker persists and the next ordinary commit is misread as an amend. And `--no-verify` is the current workaround for this very defect: adopting C would let the workaround poison the checker.

The empty-message abort (4–5) is safe — `commit-msg` runs, so the marker is consumed. Worth measuring, or it would have over-generalised to "all abort paths are unsafe".

A's loss is a check missing something; a poisoned marker is a check being **wrong**.

## Scope, and a regression the tests caught

Only `duplicate-id` ignores HEAD. My first attempt removed HEAD from the prior stream outright, which turned every `Follows:` and `Supersedes:` pointing at a record declared on HEAD into a dangling reference — two existing tests failed and named it.

## The cost, stated in SPEC rather than hidden

Re-declaring HEAD's id with a different payload on the next commit is now accepted. Measured: every other ancestor still collides (a fourth commit reusing an id from two back is still refused), and an identical payload was already permitted by §3.2.

## Verification

The amend passes and HEAD moves; the non-HEAD collision is still refused; 210 cases across `validate`, `hooks` and `schema`; `spec/verify.sh` 32 fixtures; `artifact:verify` `0d2e585b`.


MongLong0214:
Closing unmerged. This implements option A, and A is withdrawn.

## What CI showed

```
FAIL test/amend-recorded-commit.test.ts
  → a divergent record under an existing id was accepted
  → a bare id declared twice was accepted
```

Both jobs, the other eleven checks green.

That is not an implementation bug. It is the cost of A, measured — and #430 built those two assertions on purpose. Their comment says so: *"The rule this must not weaken. Two commits under one id saying different things is what SPEC §3.2 calls an identity collision, and it stays one."*

The scenario is a **new commit** re-declaring the id its parent declared, with a different payload. Its parent is HEAD. So A's stated loss and #430's guarantee are the same thing, and my estimate that the loss was "narrow" was wrong: HEAD is the position every commit passes through, which makes that the most common shape of an identity collision rather than an edge case.

## Why A looked right, so nobody proposes it again

A was chosen because the alternative that actually observes the amend — having `prepare-commit-msg` record it — appeared to fail on a measurement: `git commit --no-verify` runs `prepare-commit-msg` and skips `commit-msg`, so the marker is written and never consumed. That is true. What was not measured is what happens next: `prepare-commit-msg` runs on **every** commit path and does so before that commit's `commit-msg`, so the stale marker is overwritten before it can be read.

And the premise underneath A — *a hook cannot observe whether it is amending* — is a fact about `commit-msg`, generalised to the whole hook chain without checking. It does not hold:

```
prepare src=[commit] sha=[HEAD]   ← git commit --amend --no-edit
prepare src=[message] sha=[]      ← git commit -m
```

Git passes the amend through documented arguments. Both objections that killed option C were one measurement short.

## What replaces it

Option C, with the amend decided as `source == "commit"` **and** the sha argument actually resolving to HEAD **and** no multi-commit operation in progress — because `rebase -i` reword produces `src=[commit] sha=[HEAD]` identically to an amend and is told apart only by `rebase-merge` being present. Anything unrecognised is *not* an amend: mistaking a non-amend for an amend lets a real collision through, while the reverse is only the inconvenience this issue is about.

Tracked on #638.

Issue #667, opened 2026-08-14
duplicate-id does not catch a new commit colliding with HEAD's Record-Id


MongLong0214:
An ordinary commit that reuses the `Record-Id` already on HEAD is accepted. The
same collision is refused when the commit is an amend, so the check sees HEAD in
one path and not the other.

## Reproduction

```sh
d=$(mktemp -d); cd "$d"; git init -q .
git config user.email a@b; git config user.name a
git config --local commitlore.bin /path/to/dist/commitlore.mjs
git config --local commitlore.node "$(command -v node)"
node /path/to/dist/commitlore.mjs init

echo a > a; git add a
git commit -q -m "$(printf 'Seed\n\nLimit: one writer\nBlast: local\nUndo: easy\nCertainty: firm\nProvenance: authored\nRecord-Id: r-neg001\n')"

echo b > b; git add b
git commit -m "$(printf 'Collide\n\nLimit: one writer\nBlast: local\nUndo: easy\nCertainty: firm\nProvenance: authored\nRecord-Id: r-neg001\n')"
```

Fixture validity was asserted before reading the result — the seed landed and
HEAD carried `r-neg001` (`commits=1`).

## Observed

```
shape ok · references ok
```

## Expected

```
duplicate-id Record-Id — got "r-neg001", want "exactly one record per Record-Id"
```

## Why this matters

`Record-Id` uniqueness is what makes `Follows:` and `Supersedes:` resolvable. A
collision admitted at commit time is permanent in history, and HEAD is the
position every commit passes through, so this is the likeliest collision to
occur in practice.

Found while building a negative control for #638. The control failing is what
surfaced it: the same commit was accepted on this branch and on main, which
means #638's change does not cause it.

MongLong0214:
## Narrowed: not a timing or indexing problem

The first reading — that the index had not caught up with the commit — is wrong.
Measured on a fresh repository with two commits that both carry `Record-Id:
r-neg001`, after a full index rebuild:

```
commits=[2] both-have-id=[2]
$ commitlore init          # rebuild the index
$ commitlore validate --commit HEAD
shape ok · references ok
```

So the collision is missed **after the fact as well**, not only at commit time,
and rebuilding the index does not change the answer. Whatever the duplicate
check compares against, the other commit carrying the same id is not in it.

Both commit paths were measured and both miss it — commits made through the
hooks, and commits made with `--no-verify`.

`validate` with no arguments reports `references not checked (no repository)`
[...]

Issue #648, opened 2026-08-14
canonical-artifact.json is a derived file under version control, and it serialises every branch that touches dist


MongLong0214:
`installer/canonical-artifact.json` records checksums over all of `dist/`. Because `dist/` is tracked (274 files) and every branch rebuilds it, **any two branches that touch dist conflict on this file by construction** — not on the source change, on the manifest.

The cost is not linear. Draining N concurrent dist-touching branches forces N(N-1)/2 rebuild-and-recommit cycles, because each merge invalidates the manifest of every branch still open. With four branches that is six forced canonical rebuilds that carry no review value.

## What the manifest asserts, and who already proves it

Its fields are `format`, `builder`, `runtimeAssets`, `source`, `artifact`. `builder`, `format` and `runtimeAssets` are constants read from `scripts/canonical-artifact-contract.mjs` — source. `source.sha256` digests the declared inputs, and `artifact` digests the built bundle. Both are computed from the tree. **Every field is derivable from the commit it sits in.**

Meanwhile `.github/workflows/ci.yml` already proves the property that matters, without reference to the manifest:

```yaml
npm run build:canonical
find dist -type f -print0 | sort -z | xargs -0 sha256sum > "$first"
npm run build:canonical
find dist -type f -print0 | sort -z | xargs -0 sha256sum > "$second"
cmp "$first" "$second" || exit 1
git diff --exit-code -- dist/ installer/canonical-artifact.json
```

The canonical builder runs twice and must produce identical bytes, then the committed `dist/` must equal that rebuild. That is the integrity claim, and it needs no manifest.

## Why the committed copy cannot catch what it appears to catch

`artifact:verify` compares the committed manifest against a manifest recomputed from the tree. But the manifest is written by `npm run artifact:manifest` from whatever is in `dist/` — the same step that produced dist, in the same command. It therefore cannot detect a bad dist; it can only ever confirm itself.

The one failure it does catch is *rebuilt dist, forgot to refresh the manifest* — a workflow slip that **exists only because the manifest is committed**. The check is paying for its own existence.

## Where the manifest genuinely earns its place

At the tag. `release.yml` checks out the release commit, rebuilds canonically, and verifies — and a published statement of *which builder image, which platform, which source inputs* produced the shipped bundle is real supply-chain value a consumer can read. That value is needed **at release**, which is exactly where it is not conflicting with anything.

Nothing reads it at install time: `install.sh` and `install.ps1` do not reference it.

## Proposed shape

[...]

PR #668, opened 2026-08-14
Stop refusing an amend as a duplicate of the commit it replaces

### LINKED ISSUE #638: amend that fixes a malformed trailer is refused as a duplicate Record-Id

MongLong0214:
Closes #638.

`git commit --amend` was refused with `duplicate-id`, because the commit it
replaces is still HEAD when the record is checked and carries the same
`Record-Id`.

## How it decides

`commit-msg` cannot tell an amend from an ordinary commit — git hands it
nothing that differs. `prepare-commit-msg` can, so it writes a marker and
`commit-msg` consumes it:

    amend  =  source == "commit"
              && the sha argument resolves to HEAD
              && no multi-commit operation in progress

Anything unrecognised is not an amend. The two mistakes are not symmetric:
calling a non-amend an amend drops HEAD from the duplicate check and lets a
real identity collision through, while the reverse is only the inconvenience
this issue describes. A list of operations to *exclude* would break the day
git adds one, so the default has to be the safe direction.

Only `duplicate-id` ignores that commit. `Follows:` and `Supersedes:` still
resolve against it — removing it from the prior stream outright turned every
reference to a record declared there into a dangling one, which two existing
tests caught.

## What it does not reach

`--amend -m` and `--amend -F` supply the message on the command line, so git
reports the *message's* origin rather than the commit's and the arguments are
byte-identical to an ordinary commit. Those stay refused. This is pinned as a
test rather than left as a comment, because measured against a `-m` fixture
the working code reads as broken.

## Regressions

`test/amend-marker.test.ts` pins what git passes for each operation — ordinary
commit, amend, rebase reword, cherry-pick, revert, merge, and the `-m`/`-F`
boundary above. Reword is the case that matters: its arguments are identical
to an amend's, and only the rebase directory separates them.

Verified end to end: `--amend --no-edit` now reports `references ok` and exits
0 where it previously reported `duplicate-id`.

## Not caused by this branch

A negative control written for this work — an ordinary commit reusing HEAD's
`Record-Id` — is accepted here *and* on main. That is #667, filed separately.


MongLong0214:
## The negative control, and what this PR does not prove

A negative control was written for this work: an ordinary commit reusing the
`Record-Id` already on HEAD must still be refused. It is not. Measured on a
fresh repository that was never amended, against a build of `main`:

```
SEED-LANDED=[1] commits=[1]
shape ok · references ok
```

The fixture was asserted valid before the result was read — the seed commit
landed and HEAD carried the id.

So the control passes on `main` too, which means **it was never a valid control
for this change**, and it does not indicate a hole opened here. It is a
separate defect, filed as #667.

What that costs this PR is worth stating plainly: the safe direction — that an
unrecognised operation is not an amend, so HEAD stays in the duplicate check —
is enforced by construction and by `test/amend-marker.test.ts`, but it cannot
be demonstrated end to end until #667 is fixed, because the check it would fall
back on does not currently fire. The six-path argument table and the verified
`--amend --no-edit` pass are what this PR rests on.

## Rebase

Rebased onto `main` (was on #661, before #664 · #666 · #662 landed). The only
conflicts were the generated artifact and its manifest — #648 — resolved by
rebuilding rather than by merging: canonical artifact `a9927fa6`.

`TRACE` instrumentation used while debugging is absent from `src/` and from
`dist/`, checked on the branch rather than assumed.

The amend that produced this push was itself performed through the built CLI
carrying this fix, on a commit whose record declares `Record-Id: r-638bnd`. It
was not refused.

MongLong0214:
Correction: the comment above saying #667 blocks demonstrating this end to end is wrong. #667 was not a defect — the reproduction used byte-identical trailer payloads, which SPEC §3.2 treats as one record re-declared. A same-id collision with a *different* payload is refused (`references failed`), so the fallback this PR's safe direction relies on is present and working. #667 is closed with the measurement.

Issue #697, opened 2026-08-16
install.ps1 installs the Codex plugin and install.sh does not


MongLong0214:
Found while deleting the dead host block for #691, and it is not caused by that
deletion — the deletion revealed it.

```
install.ps1:1061   & $dest plugin install-codex        ← reached
install.sh         wire_codex_plugin { … install-codex }
                   called only from wire_codex, which nothing calls
```

On `main` today, `install.sh` contains the Codex plugin step and never runs it.
`install.ps1` runs it. **The two platform installers have been doing different
things**, and the shell's copy looked like evidence that they were not.

`test/agent-configs.test.ts` asserted the string is present in both files, which
passed — presence, not reachability. The same shape as #689: an assertion that
could not fail.

## What this means for a user

A Windows install wires the Codex plugin. A macOS or Linux install does not,
unless something else does it. Whether that gap matters depends on what the
Codex plugin provides that the MCP registration does not — the shell *does*
write Codex's MCP config, so this is about the plugin layer specifically.

## Not fixed here

#691 removes the unreachable code rather than making it reachable, because
making it reachable is a behaviour change to the shell installer and deserves its
own decision: either the shell should install the plugin (and be tested for it),
or the Windows installer should stop, and the two should be said to agree
deliberately rather than by accident.

Recording it rather than choosing, since the choice needs to know what the Codex
plugin does for a user that its MCP registration alone does not.

MongLong0214:
## The host-by-host map, so the consolidation is not guesswork

`install.ps1` both delegates and wires. Line 817 calls `installer-hosts`; from
line ~900 it does its own work. The live call sites, resolved to hosts:

```
:907   generic MCP JSON writer (Wire-McpServersJson)
:967   claude-code   installed the commitlore plugin (marketplace)
:1024  codex         registered commitlore with codex mcp add
:1045  codex         config-file branch
:1065  codex         installed the commitlore plugin (marketplace)
:1107  hermes        configured MCP + skill bundle
:1129  opencode      created the config
:1142  opencode      added into an existing config
```

Against the enumeration in `src/commands/installer-hosts.ts`:

| host | enumeration | install.ps1 | equivalent? |
|---|---|---|---|
| codex MCP | `cliHost` / `tomlHost` | :1024, :1045 | duplicated |
| codex plugin | **added in #698** | :1065 | duplicated after #698 |
| claude-code plugin | added in #690 | :967 | duplicated |
[...]

PR #701, opened 2026-08-16
docs: point the install commands at the release they install

MongLong0214:
Every install line in all four READMEs pinned v0.8.2 — three releases behind
v1.0.1. Anyone following the documented command got a build without the hook
path fix (#693), the Codex plugin on non-Windows (#697), or anything else since.

The version appears twice per command (once in the raw URL, once as the
argument), five commands per file, four files. All twenty were stale together,
which is what happens when a release moves and the documentation that names it
does not: the two are edited in different changes and only one is required.

Verified rather than assumed: both raw URLs at the v1.0.1 tag return 200, and
the installer's own usage comment names v1.0.1.


MongLong0214:
Checked before approving, and the premise does not hold on current `main`.

```
README.md       v0.8.2: 0   v1.0.1: 5
README.ko.md    v0.8.2: 0   v1.0.1: 5
README.ja.md    v0.8.2: 0   v1.0.1: 5
README.zh-CN.md v0.8.2: 0   v1.0.1: 5
```

The install commands already point at v1.0.1. They were updated in the v1.0.1
bump (#692), which moved 38 pins across nine files — 20 of them these four
READMEs, because the installers and READMEs carry the tag in URLs that only
resolve once the tag exists, so they had to move in the same commit as the
version.

This PR reports `changed_files: 0` with 2 commits, and its head branch
`docs/install-commands-v1.0.1` is not in `git ls-remote origin 'refs/heads/docs*'`.

So there is nothing to review and nothing to merge. Not closing it — it is yours,
and the discrepancy might mean the branch was created against a different base
than I am reading.

## Effect on the Reddit ordering

The reason to wait for this PR was that a tester following the documented command
would install v0.8.2 and report on the wrong build. That reason is gone: the
documented command installs v1.0.1 today.

**#700 still gates it.** A Windows tester meets exactly the case it fixes — Codex
MCP registration succeeds, the plugin step fails, and the installer reports
healthy and exits 0. A report saying "installed fine" from that state would be
describing a false success.

Issue #691, opened 2026-08-15
install.sh carries seven has_/wire_ host pairs that nothing calls


MongLong0214:
Found while fixing #689. The scope is larger than that issue described.

```
has_claude_code   called 0 times
has_codex         called 0 times
has_cursor        called 0 times
has_gemini        called 0 times
has_hermes        called 0 times
has_opencode      called 0 times
has_windsurf      called 0 times
wire_*            same, all seven
```

Every one is defined and never invoked. Roughly 140 lines. Host wiring is done by
`src/commands/installer-hosts.ts`; the shell block is what was left behind when
the work moved there.

## Why it is not merely dead

It reads as the source of truth. #689 happened because `claude-code` existed in
that block and not in the enumeration — the shell said the host was handled, so
nobody looked further. Two people read it that way today, independently, hours
apart.

Dead code that is obviously dead costs nothing. This is dead code that answers
questions.

## Interaction with #690's guard

The guard added there checks that every host defined in `install.sh` is
dispatched by the enumeration. That is useful today and rests on the shell being
a meaningful list. If the block is deleted, the guard has to become an internal
consistency check on the enumeration instead — worth doing together rather than
leaving a guard pointed at nothing.

## Recommended timing

**After v1.0.1.** A 140-line deletion is not a change to make immediately before
a release, and #690 is the urgent half — it makes the wiring exist at all.
Recording it now so it is not rediscovered a third time.

MongLong0214:
## Disposition: open, gated on the next change to `install.sh`

Not resolved by v1.0.1 — the block is still there and still answers questions.
Not urgent either: v1.0.1's controls all inverted, so there is no user-facing
reason to force a 140-line deletion now.

The gate is **the next time anyone edits `install.sh`**. That is when the cost of
reading a misleading file is actually paid, and when deleting it is one diff
rather than its own change.

## What has to move together

#690 added a guard asserting every host defined in `install.sh` is dispatched by
`src/commands/installer-hosts.ts`. It rests on the shell being a meaningful list.
Delete the block and that guard points at nothing — it must become an internal
consistency check on the enumeration in the same change, or it silently stops
guarding.

That coupling is the reason this is one piece of work rather than a cleanup
someone can do casually.

## Why it stays open rather than closing as won't-fix

It caused #689. `claude-code` was defined in the shell and absent from the
enumeration, and two people read the shell as authoritative hours apart —
independently, on the same day. Dead code that is obviously dead costs nothing;
[...]

PR #706, opened 2026-08-16
DO NOT MERGE — Run install.ps1 somewhere its behaviour can be seen

MongLong0214:
Toward the asymmetry measured on #691.

```
install.sh   → test/install-script.test.ts     4 / 4 behaviour mutations caught
install.ps1  → test/install-ps1.test.ts        0 / 5 caught
```

The POSIX tests run the script and read what it produced. The PowerShell test
reads the file as text, so a change that leaves every asserted string in place
while inverting behaviour goes unnoticed — `exit 1` → `exit 0`, a neutralised
`$LASTEXITCODE` check, three `Test-Path` typos, `ExecutionPolicy` widened, a data
root created as a file.

## The reason for that asymmetry was wrong, and it was mine

I wrote that a Linux or macOS runner cannot execute PowerShell. **This workflow
already uses `shell: pwsh` in four places.** Nobody had run `install.ps1` under
it outside Windows — that is all.

Checked before building on it: `install.ps1` uses no Windows-only cmdlets — no
registry, no `New-ItemProperty`. What it depends on is Windows *environment*
(`USERPROFILE`, `LOCALAPPDATA`) and a `.cmd` shim. The first can be injected; the
second cannot.

## What the job sees, and what it does not

Stated in the job itself rather than assumed:

- **sees** — that the installer produced diagnostics rather than exiting
  quietly, and that a directory it creates is a directory
- **does not see** — whether the `.cmd` shim executes, which is only meaningful
  to `cmd.exe`. That stays `install-ps1` on `windows-latest`.

Registered in `REQUIRED_CHECKS`: a check that cannot block a release is not a
check.

## Two conventions caught this on the way in

Both correct, and both cost a cycle: actions are pinned by SHA rather than tag,
and a new job must be declared required before the suite accepts it. The workflow
hash is re-pinned.

## What is not claimed

The 0/5 was measured against the unit test file. **Whether the existing
`install-ps1` job on Windows catches those five is unmeasured** — measuring it
means pushing five mutations through CI. So this closes a gap that exists; it
does not prove the five survive today.

91 tests passing across `release-publish-prerequisites`, `release-tag-binding`
and `action-lint`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -538,6 +538,67 @@ jobs:
   # The hosts are named literally. `shell:` accepts no expression -- a
   # `${{ matrix.shell }}` there makes the workflow unloadable, which is how this
   # job was first written and how it first failed (actions/runner#444).
+  # install.ps1's unit coverage reads the file as text, so a change that leaves
+  # every asserted string in place while inverting behaviour goes unnoticed:
+  # measured at 0 of 5 such mutations caught, against 4 of 4 for install.sh,
+  # whose tests run the script and read what it produced (#691).
+  #
+  # The reason given for that asymmetry was that a Linux runner cannot execute
+  # PowerShell. It can — this workflow already uses `shell: pwsh` in four
+  # places. Nobody had run install.ps1 under it outside Windows.
+  #
+  # What this job can and cannot see is worth stating: the script's Windows
+  # environment variables are injected, so its path logic runs, but the .cmd
+  # shim it writes is only meaningful to cmd.exe. Whether that shim *executes*
+  # stays a windows-latest question, and install-ps1 below remains the job that
+  # answers it. This one answers the cheaper half, on every push.
+  install-ps1-behaviour:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Run install.ps1 under PowerShell Core with a synthetic Windows home
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+          $home_ = Join-Path ([System.IO.Path]::GetTempPath()) "cl-home-$(Get-Random)"
+          $local = Join-Path $home_ 'AppData/Local'
+          New-Item -ItemType Directory -Force -Path $local | Out-Null
+          $env:USERPROFILE = $home_
+          $env:LOCALAPPDATA = $local
+
+          # A failure here is the point: the assertions below say which part.
+          & ./install.ps1 2>&1 | Tee-Object -Variable output | Out-Host
+          $code = $LASTEXITCODE
+          "$output" | Set-Content -LiteralPath "$home_/output.txt"
+          "exit=$code" | Out-Host
+          "$code" | Set-Content -LiteralPath "$home_/exit.txt"
+          "$home_" | Set-Content -LiteralPath "$env:GITHUB_ENV" -Append -Value "CL_HOME=$home_"
+
+      - name: It reports what it did, rather than exiting quietly
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+          $out = Get-Content -Raw -LiteralPath "$env:CL_HOME/output.txt"
+          # Catches `exit 1` -> `exit 0` and a neutralised $LASTEXITCODE check:
+          # both make the script claim success while having done nothing.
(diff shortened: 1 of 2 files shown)


MongLong0214:
**Do not merge.** Measured in #707: this job's first step exits 1 before anything is asserted.

`install.ps1` returns non-zero on Linux — it cannot write a usable `.cmd` shim — and `$ErrorActionPreference = 'Stop'` turns that into a step failure. Since the job is registered in `REQUIRED_CHECKS`, merging it makes a required check permanently red and blocks releases.

Three claims of mine were wrong, each one under the last:

1. *a Linux runner cannot execute PowerShell* — false; this workflow already uses `shell: pwsh` in four places
2. the `# Catches X` comments — written without running X
3. that the job runs at all — it does not

## What this needs before it is mergeable

1. Tolerate a non-zero exit as an **observation** rather than a step failure — the exit code is data here, not an error.
2. Extract the three assertions into `Scripts/assert-install-ps1-behaviour.ps1` returning 0/1, and have both the job and any measurement call **that**. The measurement in #707 reimplemented two of the three, so its number would have described a twin rather than this job — which is the same defect this PR exists to fix.
3. Only then measure N/5, and record it here and in the job's comments.

Step 2 is the review's point and is right; step 1 has to come first or step 2 changes nothing.

MongLong0214:
Closing. Measured, and the premise does not survive the measurement.

## The job this PR duplicates already runs on every push

```
.github/workflows/ci.yml:542        install-ps1  runs-on: windows-latest   (no `if:`)
scripts/check-exact-head-ci.mjs:55  'install-ps1'                          already in REQUIRED_CHECKS
ci.yml:4-6                          push: [main, dev, release-*, hotfix-*] + pull_request
```

My stated justification was that this job answers *the cheaper half, on every push*. That half was already answered on every push, on a real Windows runner, by a job that asserts more: shim presence, the `commitlore:wrapper:v1` marker, CRLF, declared-version agreement, exit 3 for a wrong tag, byte-identity of the shim after a refused install, and the repair command extracted from the log and executed. This PR's three assertions are a subset.

## Measurement against that job

Four throwaway `hotfix-*` branches, one mutation each, reading the `install-ps1` job's step conclusions — no harness, no copy of the assertions.

| mutation | sites | `install-ps1` | failing step |
|---|---|---|---|
| `$LASTEXITCODE -eq 0` → `$true` | 6 | **MISSED** | — |
| `Test-Path` ×3 misspelled | 3 | CAUGHT | Install under Windows PowerShell 5.1 |
| `-ItemType Directory` → `File` | 6 | CAUGHT | Install under Windows PowerShell 5.1 |

**RESULT 2 / 3.**

Two of the original five were never mutations, which is why the denominator is 3:

- `ExecutionPolicy Bypass` → `Unrestricted` — `Bypass` appears **0 times** in `install.ps1`. It lives in CI's `powershell.exe` invocation.
- `exit 1` → `exit 0` — the one occurrence of `exit 1` is inside a **comment** (`install.ps1:304`). The script's real exits are 0, 2, 3, 4.

Both would have counted as MISSED and pushed the score down, which reads as *more coverage needed* — the conclusion this PR was built on.

## What the two catches actually were

Both failed at **Install under Windows PowerShell 5.1** — the script crashed while installing. Every downstream assertion step was `skipped`. So the job's assertions discriminated **0 of 3**; execution did the work. For `-ItemType File` the runner shows exactly why:

```
commitlore-install: installing v0.8.2
commitlore-install: error: could not fetch v0.8.2 ...
fatal: could not create leading directories of
  'C:\Users\runneradmin\AppData\Local\commitlore\.v0.8.2.incoming.7824': Not a directory
Nothing was installed.
```

## Why this PR does not close the one real gap

The surviving mutation forces six wiring outcomes (`$gitRan`, `$added`, `$ok`, `$registered`, `$hermesOk`) unconditionally true. This job's assertions are: diagnostics present, no `installed to` without a name, nothing named `commitlore` created as a file. A wired-nothing install still prints diagnostics and still creates directories — so this job would report green too. It adds a required check that misses the only defect the measurement found, on a runner where the `.cmd` shim it would need is meaningless.

Evidence: [m2](https://github.com/MongLong0214/commitlore/actions/runs/31957479583) · [m3](https://github.com/MongLong0214/commitlore/actions/runs/31957479737) · [m5](https://github.com/MongLong0214/commitlore/actions/runs/31957479120) · [m1 — invalid](https://github.com/MongLong0214/commitlore/actions/runs/31957479101)

The gap is recorded on #691, where the installer wiring work lives.

PR #707, opened 2026-08-16
MEASUREMENT ONLY — do not merge

MongLong0214:
Temporary. Runs the five behaviour mutations against the new install-ps1-behaviour job in one CI run, to replace the unmeasured 'Catches X' claims in #706 with N/5. Closed as soon as the number is read.


MongLong0214:
Closing: the harness never ran. The base step exited 1 before the measurement loop — install.ps1 exits non-zero on Linux (it cannot write a usable .cmd shim) and `$ErrorActionPreference = 'Stop'` turns that into a step failure.

So there is no N/5 from this. The twin problem raised in review is real and unaddressed on top of that: the loop reimplemented two of the job's three assertions instead of calling the job's own.

Issue #716, opened 2026-08-17
Windows: every detected host fails to wire — the temp filename carries the whole path, and hasCommand cannot see a .cmd


MongLong0214:
An owner ran v1.0.2 on Windows 10.0.19045.0 with agents present. Every detected host failed to wire; nothing was written to any config. Full observation and what it settles: #714.

The installer reported the failure honestly — `ok:false`, per-host `outcome:"failed"`, configs byte-identical to their backups. It did not claim work it had not done. What it could not do is the work.

Two causes are visible in `src/commands/installer-hosts.ts` without a Windows machine.

## 1. The temporary filename is built from the whole path

Both atomic writers (the JSON one and the Codex TOML one):

```ts
const temporary = join(dirname(path), `.${String(path.split('/').pop())}.commitlore-${process.pid}-${randomUUID()}.tmp`);
```

`split('/')` does not know the Windows separator. On `C:\Users\u\.gemini\settings.json` there is no `/` at all, so `.pop()` returns the entire string and it is used as a filename:

```
open 'C:\Users\u\.gemini\.C:\Users\u\.gemini\settings.json.commitlore-41756-….tmp'
      └── dirname ──────┘  └── the whole absolute path, as a name ─────┘
```

A drive letter cannot appear inside a filename, so this is ENOENT every time. It is invisible on POSIX because `split('/')` and `basename` agree there.

`basename(path)` is the fix, and `node:path` is already imported. Every JSON host that reaches its write is affected, not only gemini-cli.

## 2. `hasCommand` cannot find a Windows executable

```ts
const hasCommand = (command: string): boolean =>
  (process.env.PATH ?? '').split(delimiter).some((directory) => {
    const path = join(directory, command);
    try { return !statSync(path).isDirectory(); } catch { return false; }
  });
```

Windows installs `cursor.cmd`, `codex.cmd`, `claude.cmd`; there is no extensionless file, and `PATHEXT` is never consulted. Consistent with the run: `claude-code` reported `notDetected` while its config file was present on disk.

The effect is not only detection. Where a name *is* found — the tester's PATH did have something called `codex` — the command is then run with:

```ts
spawnSync(command, args, { encoding: 'utf8', shell: false })
```

[...]

Issue #714, opened 2026-08-17
Verified, and it fails: installer-hosts wires nothing on a real Windows machine


MongLong0214:
> **This is a record, not a work item.** Nothing is scheduled against it and nobody is waiting on a machine. It exists so that the unverified part of a shipped claim stays written down instead of disappearing when #691 closed.
>
> If a Windows user reports that hosts are not wired, this is the difference between a narrow investigation and starting from nothing.

Carries forward the one claim #691 left open, and the concern #712 was filed about. Both are closed; this is not.

## What is unverified

`installer-hosts` detects coding agents, writes their configs and probes the result. On Windows nobody has ever watched it do that with agents present.

| platform | status |
|---|---|
| macOS | verified on a real machine — the v1.0.2 readback showed `codex … plugin installed` and `claude-code` in `hosts` |
| Windows | **never run with an agent installed** |
| Linux / alpine | not applicable — the installers there wire no agent hosts |

## CI cannot answer this, structurally

A GitHub runner has no coding agents, so every host short-circuits at detection. Measured, printed by the `install-ps1` job on every run:

```json
{"schema":"commitlore_installer_hosts.v1","ok":true,
 "hosts":[],
 "notDetected":["codex","gemini-cli","cursor","windsurf","opencode","hermes","claude-code"]}
```

**This is the part to read before proposing a fix.** An assertion added to CI would read an empty list and pass forever — a check that cannot fail is worse than a missing one, and this repository has spent real time removing exactly that shape (#700, and the guard rewritten in #713). Adding a stub host to make detection succeed has the same problem in a new place unless the stub can *fail*: a stub that always registers cleanly proves the installer reports what the stub told it.

## Why it got sharper, not weaker, after #713

#691's step 2 was gated on this machine because `install.ps1` appeared to wire hosts itself, so the question was whether deleting that changed the result. It could not — the code sat after an unconditional `exit`. #713 deletes it.

What that leaves is a single path. Before #713 one could imagine the shell copy as a fallback; there is no copy now. **The never-verified path is the only path.**

## What would close this

One Windows machine with at least one agent installed, one install, and the `hosts` array read afterwards. Whether a failing registration is reported as failed rather than wired is the second half, and the more valuable one.

Same machine that #691 wanted; a narrower and now clearly stated question.

MongLong0214:
Confirmed by the owner, 2026-08-17: **Windows verification is not going to be done.** This issue stays open as the record of what that leaves unverified, not as work.

[...]

PR #717, opened 2026-08-17
Stop building the temporary filename out of the whole path (#716)

MongLong0214:
Partially addresses #716; the observation that produced it is on #714.

An owner ran v1.0.2 on a real Windows machine with agents present — the run #714 was holding open. Every detected host failed and no config was written. gemini-cli printed the cause:

```
open 'C:\Users\u\.gemini\.C:\Users\u\.gemini\settings.json.commitlore-41756-….tmp'
      └── dirname ──────┘  └── the whole path, used as a filename ────┘
```

Both atomic writers built the temporary name with `path.split('/').pop()`, which returns its argument unchanged when there is no `/` in it — true of every Windows absolute path. A drive letter cannot be part of a filename, so the write was ENOENT before it started.

## Why not `basename`

`basename` is correct here on Windows. It is also unfalsifiable anywhere else, and this repository has no Windows evidence to lean on: CI has no coding agent, so the `install-ps1` job detects nothing and never reaches this line — the whole of #714. Splitting on either separator makes the property one a POSIX runner can fail on, which is the only kind of guard that can hold a Windows-only defect here.

## Not fixed here

The second cause in #716: `hasCommand` stats each PATH entry with the bare command name, so it never finds `cursor.cmd` or `claude.cmd`, and `spawnSync(..., { shell: false })` cannot execute a `.cmd` shim. That accounts for `codex mcp add failed`, `Hermes setup failed`, and `claude-code` reported `notDetected` with its config present. Fixing it needs a design that preserves the reason `shell: false` was chosen — the wrapper path and config paths reach those calls — so it is not folded in here.

## Verification

Restoring `split('/')` fails 2 of the 4 new cases, including the Windows path producing `.C:\Users\u\.gemini\settings.json.commitlore-pid-uuid.tmp` where `.settings.json.commitlore-pid-uuid.tmp` is required. 7 cases pass across the two `installer-hosts` suites; `tsc --noEmit` clean; `build:canonical` then `artifact:verify` → `8613363cdee2183e7a6f693b31450d715f3f316ba32a4ed77272b8af9d9c9c0f`.

**No Windows machine has run this fix.** The guard proves the name, not the install. #716 stays open for the second cause and for a Windows readback.


MongLong0214:
Post-merge note on the artifact digest.

This branch's record (`r-wintmp716`) verifies `artifact:verify -> 8613363cdee2183e7a6f693b31450d715f3f316ba32a4ed77272b8af9d9c9c0f`. That is true of the commit it sits on. After #715 landed, `main` moved and this branch merged it; the two committed build artifacts conflicted, which is expected — three open PRs each regenerate `dist/commitlore.mjs` and `installer/canonical-artifact.json`.

Resolved by regenerating rather than by picking a side. A generated file has no correct side; the only answer is to rebuild it from the merged source:

```
build:canonical then artifact:verify -> d20d47b227b226b1b352fc5c6f9527d9b1653115d54536196948b79fdbfe3a6f
```

So a reader checking the record against the merged tree finds two different digests, and both are right about different commits. Recorded here so that difference is not read as a build that drifted.

No force push was used — the update is a merge of `origin/main` into this branch.

Issue #742, opened 2026-08-17
No way to learn a newer release exists — a stale install quietly enforces an older protocol


MongLong0214:
There is no way to find out that a newer CommitLore exists. `--version` reports what is running and nothing compares it to anything.

That matters more here than for most tools, because of how this one is installed and what a stale install does.

## Why a stale install is quiet

`commitlore init` records the hook's interpreter as `<data-root>/current/dist/commitlore.mjs`. A repository keeps validating commits with whatever `current` points at, and nothing in a normal day tells the operator which release that is. `doctor` reports the split when asked — that is how #735 was caught — but nobody runs `doctor` on a machine that appears to work.

So an upgrade nobody knows about is not a missing feature, it is a repository quietly enforcing an older protocol.

## The shape, and why it is not self-replacement

Two pieces, matching what `gh`, `npm` and `rustup` do:

**1. A passive notice.** After a command, if a newer release exists, print one line naming both versions and the command to upgrade. Cached so the network is touched at most once a day, silent on any failure, off in CI and on a non-TTY, and switchable off entirely. A notice that costs latency or noise gets suppressed by the user, and then it is worse than nothing.

**2. `commitlore update`.** Explicit. Reports current, latest, and the exact upgrade command — and by default **does not run it**.

That last part is the decision worth arguing. A tool that replaces its own bytes without being asked is a supply-chain step taken on the user's behalf, and `gh` deliberately does not do it. Neither should this: the install path is `install.sh` fetching a pinned tag, and the CLI reimplementing that would be a second implementation of the one thing #691 spent 845 lines proving should have exactly one. The installer already verifies its own work after #735 — it reads `current` back rather than trusting an exit code — and duplicating it would duplicate that too, or worse, not.

So `update` prints the command that runs the installer, the same one the README gives.

## What it must not do

- **Report a version it did not check.** A cached answer says when it was cached.
- **Fail a command because the network failed.** The notice is best-effort; the command's exit code is about the command.
- **Speak in CI.** Every CI run printing an upgrade notice trains people to ignore the line that matters.
- **Compare wrongly.** `1.1.10` is newer than `1.1.9`; a string comparison says otherwise.

## Open

Whether the notice belongs on every command or only on `doctor` and `init`. Every command is what other tools do; this one has hook-invoked subcommands (`prepare-commit-msg`, `post-commit`, `pre-push`) where any output is either noise or a corrupted commit message, and those must be silent regardless.

MongLong0214:
[...]

Issue #735, opened 2026-08-17
install.sh reports `current -> vX` while leaving `current` on the previous version


MongLong0214:
## The installer reports `current -> vX` and leaves `current` pointing at the previous version

Upgrading to 1.1.1 on macOS printed the success line and did not move the symlink.

```
commitlore-install: checked out and verified v1.1.1 into /Users/isaac/.local/share/commitlore/v1.1.1
commitlore-install: current -> v1.1.1                       <- claim
commitlore-install: installed to /Users/isaac/.local/bin/commitlore
1.1.1
```

Afterwards:

```
$ readlink ~/.local/share/commitlore/current
/Users/isaac/.local/share/commitlore/v1.0.2

$ node ~/.local/share/commitlore/current/dist/commitlore.mjs --version
1.0.2
```

## Why it matters

`commitlore init` records the hook's interpreter as `<data-root>/current/dist/commitlore.mjs`, which is what makes hooks follow upgrades instead of going stale. When `current` does not move, every repository on the machine keeps validating commits with the old build while the CLI reports the new one.

`doctor` catches the split and says so clearly:

```
warn  commit-msg hook -- commitlore.bin is version 1.0.2, but this CLI is 1.1.1 --
      the hook validates every commit with 1.0.2, so anything fixed since then does not apply here
warn  runtime identity -- hook v1.0.2 ... CLI v1.1.1
```

So the state is recoverable. What is not is the installer's report: it said it had done the thing, and the operator has no reason to run `doctor` after a success line.

## Cause

`install.sh:645-648`:

```sh
current_link="$data_root/current"
current_tmp="$current_link.commitlore-install.$$"
if ln -sfn "$candidate" "$current_tmp" 2>/dev/null && mv -f "$current_tmp" "$current_link" 2>/dev/null; then
  log "current -> $(basename "$candidate")"
```

`$current_link` already exists and is **a symlink to a directory**. BSD `mv` follows such a destination and moves the source *into* it, rather than replacing it. It succeeds, so the `&&` holds and the success line prints.

The temp link is still on disk where `mv` actually put it:

```
$ find ~/.local/share/commitlore -maxdepth 2 -name 'current*'
/Users/isaac/.local/share/commitlore/current                                    -> v1.0.2   (untouched)
/Users/isaac/.local/share/commitlore/v1.0.2/current.commitlore-install.53786    -> v1.1.1   (where mv put it)
```

[...]

PR #744, opened 2026-08-17
Specify release awareness against read sources, not remembered ones (#742)

MongLong0214:
PRD, ADR and tickets for #742. This is revision 2 — revision 1 argued from convention and got the convention backwards, so every claim below was read at the source before it was written down.

## What the references overturned

| claim in revision 1 | what the source says |
|---|---|
| `rustup` supports "does not replace itself" | **backwards.** `auto-self-update` is enabled by default; `--no-self-update` exists because self-update is the default |
| convention is notify-only | **three of six self-replace** — `rustup`, `deno upgrade`, `bun upgrade`. Only `gh` and the `update-notifier` family notify |
| "the releases API is what the install path already uses" | **false about this repository.** `install.sh` has zero `api.github.com` calls; it uses `git ls-remote --tags --refs`, above a comment saying that needs "no API token and no rate limit" |

That last one mattered most: the API was chosen on a false premise, and choosing it would have **regressed to a limit the installer had already escaped** — 60 requests/hour/IP, which a fleet behind one NAT reaches on a cold cache. The check now uses the same `git ls-remote`, which also works where `api.github.com` is blocked but git is not, and honours `COMMITLORE_INSTALL_SOURCE` for free.

## What changed in the design

**ADR-0037 is narrowed.** The `--apply` rejection is withdrawn — it contradicted the ADR's own sentence that *"the right shape is for the CLI to invoke the installer, not to reimplement it"*, which describes `--apply`. What survives is narrower and load-bearing: **no second installer in TypeScript**, on #691 (one implementation) and #735 (that implementation verifies its own work). What replaces the rejection is an obstacle that was not visible before — `<data-root>/current/install.sh` is the **old** installer, so applying an upgrade with it inherits its defects, and #735 is the proof rather than the hypothetical.

**The suppression rules are split in two.** Revision 1 gave the passive notice and `commitlore update` one table, so a nightly `commitlore update --check` — `CI` set, no terminal, by construction — would have returned a silent "unknown" from the one command whose purpose is to be scripted. Context gates (`CI`, TTY, hook subcommand) belong to the uninvited line; a typed command obeys only the explicit off-switches.

**`--check` exits 0 either way**, following `src/commands/auto.ts:16` — *"the answer is not a finding"*. Scripts branch on `--json`, as they do with `doctor`.

## Two things reading `gh` changed rather than confirmed

- It runs the check **concurrently and cancels it when the command ends** (`cmd.go:143-152`, `cmd.go:255`), so a slow network costs nothing by construction rather than by a tuned timeout.
- It writes its state **only after a successful fetch** (`update.go:103`), so an offline machine retries on every invocation forever. We cache the failure too — an outbound attempt per command is exactly what the air-gapped case is trying to avoid.

Also added: `DO_NOT_TRACK` (its stated scope names autoupdates), and a **config-file** off-switch alongside the env vars, because an org can ship a file to a fleet but cannot guarantee an env var in every shell — Terraform's `disable_checkpoint` split.

## Scope

Specification only. No code. T-1601 → T-1604 are ordered strictly, and every acceptance criterion is unproven until T-1601 lands.


MongLong0214:
## Revision 3 — a second review round refuted all four remaining claims

Round 2 went after the claims **this PR introduced**, which is where the value was: three broke, and one of them reverses a decision.

### `--apply` is restored, and #742's request is no longer narrowed

Revision 2 replaced a bad rejection with an obstacle: `<data-root>/current/install.sh` is the **old** installer, so applying an upgrade with it inherits its defects — #735 as proof rather than hypothesis. That is checkable, and checking it broke it:

```
install.sh:388   version="${1:-}"                              a version arg skips discovery
install.sh:548   git clone --depth 1 --branch "$version"       the clone never touches `current`
install.sh:523   reuses an existing checkout (manifest + tag verified)
install.sh:669   readlink -- the #735 repair reports from the link, not the exit code
```

#735's only observable symptom is a `current` that did not move while the exit code said otherwise — **detectable from outside by reading the link.** So ADR-0038 specifies four steps: call the on-disk installer for the target tag, read the link back, and on failure call the **newly downloaded tree's** installer, which carries the fixed move and reuses the checkout phase one just made.

Reading a symlink is not installing, so ADR-0037's core is untouched: still no second implementation.

**ADR-0037 was wrong twice in the same direction — narrowing a request the owner made.** #742 asked for a version check *and* an update capability. Both rejections were reasoned rather than measured, and both dissolved when the source was read.

### `doctor` was in neither suppression class, and the gap had teeth

T-1604 silences every `--json` invocation, while the PRD said scripts branch on `--json` "which is what `doctor` does". So **the one structured contract anybody consumes could not carry staleness, and the notice could not appear there either** — invisible in exactly the output built for programs to read. `doctor` is now the third class (T-1605): a finding in its own report and a field in `doctor --json`, ignoring TTY and `CI` like a command and unlike a notice.

### Failure caching was right in kind, wrong in dose

One 24-hour interval for every failure turns a five-minute outage into a day of silence about the staleness this feature exists to expose. Split: unreachable backs off from an hour, a refusal waits a day, output-no-tag-matched is a parsing bug that gets logged rather than buried. And the air-gapped case never justified it — the explicit off-switches already serve that, fleet-wide.

### Cancellation was an outcome, not a mechanism

"A test asserts the process is gone" is satisfiable two ways with only one of them right. Now named: process group, a hard timeout **independent of the caller** (`commitlore --version` finishes in milliseconds, so cancel-on-completion is not a bound), `SIGTERM` then `SIGKILL`, reaping, `GIT_TERMINAL_PROMPT=0`. The exposed surface is every command that is **not** a hook subcommand — hooks already have the check off.

---

Tickets are now T-1601 → T-1602 → T-1603 → { T-1604, T-1605 } → T-1606, and nothing changes the machine until everything above it is right.

PR #752, opened 2026-08-18
Replace the competing README visuals with one lifecycle diagram

MongLong0214:
Two SVGs sat above the fold in incompatible art directions — a warm editorial diagram and an animated dark terminal recording — so the first screen read as two templates rather than one product. The recording is retired and the hero replaced.

## Reference search, before and after

```
$ git grep -n -e 'commitlore-demo\.svg' -e 'record-demo\.mjs' -e 'T-1016'
before   4 READMEs · the asset · its generator · its byte-identity test · 3 historical ticket/gate docs
after    3 historical ticket/gate docs only
```

`commitlore demo`, its scenario and its tests are untouched. What is deleted is the README recording asset, the generator that existed only to produce it, and that generator's test.

## Three departures from the reference skeleton

Each because following it exactly would have shipped something worse.

**The skeleton put `SUPERSEDED` in the top slot**, and §5.2 says the superseded row must not take the eye first. The hero being replaced already had the active row on top:

```
old hero   ACTIVE y=80    SUPERSEDED y=256    satisfies the rule
skeleton   SUPERSEDED     ACTIVE              inverts it
```

Copying the skeleton would have broken a property that already held. Newest-first also matches what `git log` shows, so a column labelled `GIT HISTORY` now reads the way the tool it names does. **This is a defect in the handoff, not a preference** — §5.2 and §7.7 contradict each other.

**The next-edit column mixed two x positions for one phrase** (690 / 674 / 674) and carried a green dot competing with the arrow that already supplies the accent — a fourth green mark in the region §5.2 ranks fourth for attention. The dot is gone under §6's "no decorative illustration" rather than under "the forbidden list does not mention it", which is the weaker argument. Its removal is why the three lines needed realigning at all.

**Panel padding is 16 rather than 22 on both sides.** `Checkout pricing only` at 24px lands within ~2px of the old 284px limit on a wide system font, and this stack resolves to whatever the reader's OS provides — so the margin was thin on somebody else's machine, not on mine. Within §12's `spacing ±6px`, and symmetric.

## The alt text was making a claim the product does not

```
- alt="CommitLore: a coding agent must not revive a decision the repository already reversed."
+ alt="CommitLore delivers the active decision for src/pricing.ts before the next edit, while a
+      superseded decision does not move forward as current guidance."
```

All four translations carried the same claim; all four are corrected. CommitLore does not deliver a superseded record as current guidance — it does not stop an agent from thinking of the idea again.

## Negative controls

`test/readme-hero.test.ts` is new. Each mutation was applied, the suite run, then restored:

| mutation | result |
|---|---|
| restore the demo block to `README.md` | 2 failed |
| add `<set>` to the hero | 1 failed |
| replace the alt with `the agent cannot revive it` | 1 failed |
| delete `<desc>` | 1 failed |
| put the superseded row back on top | 1 failed |
| restore | 24 passed |

The first version of the "no absolute claim" test failed on a source comment quoting the design rule it was implementing. The test now reads rendered text plus the accessible name and description — what actually reaches a reader — rather than the whole file.

## Unresolved, and stated rather than hidden

**375px.** The README embeds at `width="100%"`, so a 375px viewport scales the 840px canvas by 0.446:

```
18px → 8.0px    GIT HISTORY · SUPERSEDED · ACTIVE · CURRENT CONTEXT · Limit · Ruled out · NEXT EDIT
24px → 10.7px   the largest text in the diagram
                    ↑ a caption usually needs 12 CSS px; Apple HIG says 11pt
```

Better than the 1200px canvas it replaces (7.5px) and still short of the floor. §18.2 makes those exact labels a pass criterion, and §9.2's 18px minimum does not reach it — **24px does not either**. Three columns and this copy do not fit under a 700px canvas, so the canvas question is deferred rather than answered, and the measurement is recorded in the test file so it is not rediscovered.

Also unresolved: §12's allowed-changes list cannot express the fix for the overflow above — font reduction, line break and panel widening are all outside it.

## Not changed

`commitlore demo` CLI, lifecycle behaviour, capture, protocol, runtime.


MongLong0214:
The commits from this branch are in `main` — `8dcce526` is an ancestor of the 1.1.3 merge commit `e3ee9e8`, landed by #760 with `--no-ff`.

GitHub recorded this as *closed* rather than *merged*, and the reason is worth writing down: #760's body contained the sentence "GitHub closes #752, #755, #756, #757, #758 and #759 as merged". A closing keyword binds only to the number immediately after it, so `closes #752` registered as a linked reference and closed this one by keyword, while the other five were detected by ancestry and marked merged.

The sentence explaining that they would close as merged is what stopped one of them from doing it. The label is wrong; the code is where it should be.

Issue #723, opened 2026-08-17
A fork PR's records are discarded on squash: preserve cannot push notes, and it is not a required check


MongLong0214:
A pull request from a fork cannot publish `refs/notes/commitlore`, so squash-merging one **discards every record in it**. Observed on #720:

```
merge-type=squash  records=4  conflicts=0  pushed=false

ERROR: could not publish refs/notes/commitlore to origin after 1 attempt(s):
  remote: Permission to MongLong0214/commitlore.git denied to github-actions[bot].
  → the record was attached on the runner and is about to be discarded with it
```

GitHub gives a fork PR's workflow a read-only token. `contents: write` in the workflow cannot grant what the event does not carry, so the note is built correctly on the runner and then thrown away with it.

## The action behaves well; the repository does not protect it

The action fails loudly, names the cause, and says exactly what is about to be lost — *"the record was attached on the runner and is about to be discarded with it"*. That is the right report.

**But `preserve` is not one of the eleven required status checks on `main`.** So a red `preserve` does not block the merge button. The failure mode is:

1. a fork PR carries records in its commits;
2. `preserve` goes red for a reason that reads like a permissions nuisance;
3. every required check is green, so the PR merges;
4. the squash collapses the commits and the records are gone, with nothing left to notice.

This is the thing this project exists to prevent, reached through the one path where its own guard has no authority.

## What it does not affect

A pull request from a branch in this repository. Those get a writable token and `preserve` works — which is why this has never been seen before. The gap opens exactly when a contributor works from a fork, and #720 is the first one.

## Options, none of them obviously right

- **Merge commits for fork PRs.** The records stay in the commit messages, which is their primary home; notes are the mirror for squashed history. Costs the squashed history the project otherwise keeps. This is what #720 will use.
- **Make `preserve` a required check.** Then a fork PR simply cannot merge until someone intervenes, which is honest but blocks every outside contribution on a permissions fact they cannot fix.
- **`pull_request_target`**, which does carry a writable token — and runs workflow code from the base branch against a fork's content, which is a well-known way to hand an attacker a token. Named here so it is rejected on the record rather than proposed later.
- **A maintainer runs `commitlore squash-preserve` after merging.** The tool exists for exactly this. It is also a manual step nobody is reminded of, which is the shape that fails on the fifth occurrence rather than the first.

## What a fix has to keep

[...]

Issue #722, opened 2026-08-17
Plant a host fixture in Windows CI so this class can't hide behind an empty runner


MongLong0214:
> **So this class can't hide behind an empty runner again.** The phrasing, and the idea, come from a public suggestion by kantorcodes1: plant a fake `cursor.cmd` on `PATH` in the Windows job and assert detection and wiring through the same `installer-hosts.ts` path.

## The problem, stated as a property of the runner rather than of the code

A GitHub runner has no coding agents installed. So every host short-circuits at detection, and the `install-ps1` job prints this on every run:

```json
{"schema":"commitlore_installer_hosts.v1","ok":true,
 "hosts":[],
 "notDetected":["codex","gemini-cli","cursor","windsurf","opencode","hermes","claude-code"]}
```

Every branch that wires a host is therefore unexecuted in CI, always. Not under-tested — **unreachable**. That is why #714 exists as a standing record rather than a task, and why the Windows defects in #716 were found by an owner on a real machine and not by any job here.

The response so far has been to make the *guards* falsifiable somewhere the code does run: #717 pinned the temporary-filename property as a pure function a POSIX runner can fail on, and #720 added a mode-0644 shadow test for executable resolution. That works, and it is second best. It proves properties of helpers. It does not execute the path.

**The better answer is to stop accepting the empty runner as a constraint.** One planted host makes the branch run, and the evidence changes from an argument about control flow into an execution.

## What this would have caught

| defect | how it was actually found | with a planted host |
|---|---|---|
| #716 cause 2 — `hasCommand` never consults `PATHEXT`, so a `.cmd` is invisible | an owner installing v1.0.2 on a real Windows machine | the fixture is a `.cmd`; detection fails in CI |
| #716 cause 1 — the temporary filename carried the whole path | the same real machine, as `ENOENT` | the write is reached, so the `ENOENT` is reached |
| the X_OK blind spot — `wrapper()` in `test/installer-hosts.test.ts` hard-codes `chmodSync(path, 0o755)`, so no fixture can be non-executable | reading the helper while reviewing #720 | a real-path fixture chooses its own mode |

Three things that were being tracked separately. They are one thing: **a check whose subject is not present.**

## Why it has to be able to fail

A stub that always registers cleanly proves that the installer reports what the stub told it, and nothing else — the same shape as an assertion that reads `notDetected: []` and calls it success. So the fixture has to support at least:

[...]

PR #754, opened 2026-08-18
Build the commit that lands: canonical merge as one pull request (T-1502, #719)

MongLong0214:
**Rebuilt after review.** The first shape let a wrong commit reach `main` and opened a second pull request to fix it. That does not meet T-1502's acceptance — *"a source-only pull request merges and the commit that lands on `main` passes `artifact:verify` and `git diff --exit-code -- dist/` without anyone rebuilding by hand"* — because the commit that lands is the wrong one, briefly, every time. It was also building a file T-1502 owns while T-1501 puts *"any workflow file"* out of scope.

## What it does now

`main` + the pull request's source + a canonical rebuild, on one branch, as one pull request. **All eleven required contexts run on the tree that actually merges**, rather than on one that resembles it. A contributor who cannot produce a `linux/amd64` Docker build waits for nothing — which is #720.

## The safety property with no precedent in this repository

The rebuild runs `npm ci` and `build:canonical` **on the merged tree**, which executes whatever the pull request put in `package.json`. So:

```
1  fetch the PR into a local ref            nothing executed
2  refuse anything not source-only          before any execution
3  merge onto main
4  npm ci + build:canonical                 ← contributor code runs here
5  refuse if main moved                       ← NO App token in the environment yet
6  mint the App token                       ← only now
7  push, open the pull request
```

**The order of steps 4 and 6 is the whole guard**, and `test/canonical-merge-workflow.test.ts` asserts it with comment lines stripped, so the workflow's own explanation cannot satisfy the check (#723's shape).

## Three findings from the review, fixed rather than argued with

**The concurrency group orders jobs, not merges.** My own comment claimed it removed a race and then described that race. A rebuild is of `main` as it stood when the job started, and nothing stops another pull request landing while it runs. There is now a check between the rebuild and the push.

**A rebuild branch cannot be updated, only replaced.** `main` is `strict`, and `.gitattributes` marks `dist/**` as `-merge`, so "Update branch" conflicts on the one file this job exists to produce. The branch is keyed to the pull request rather than to a commit, and a rerun force-pushes.

**The guard was blind on merge commits.** `git show --name-only` prints a merge's *combined* diff — only paths differing from every parent — so a clean merge reports nothing:

```
5dcc02b (a real merge on main)
  git show --name-only                0 paths   ← the first draft read this
  git show --first-parent --name-only 7 paths   ← what it reads now
  diff-tree --first-parent            0 paths   ← the suggested replacement; measured, not adopted
```

## Negative controls

| mutation | result |
|---|---|
| move the token mint above the rebuild | 2 failed |
| delete the source-only refusal | 1 failed |
| restore | 11 passed |

Plus `test/canonical-rebuild-guard.test.ts`: five cases, each a real repository with a real commit, and removing `--first-parent` fails the merge case. 79 tests across five workflow suites.

## Run by hand, deliberately

An automatic trigger would let a fork's push decide when this repository's App token sits in an environment building that fork's code. #723 is the record of why a workflow must not take direction from a pull request head.

## Limit

Pull requests can still carry `dist/` when they choose to. **T-1503** is what stops requiring it; until then a source-only pull request is a contributor's option rather than the rule, and #719's success line ("no pull request carries `dist/`") is not reached by this alone.


MongLong0214:
**Holding this out of the 1.1.3 release**, on its own ticket's requirement rather than on caution.

`T-1503`'s dependency line:

> **Depends on** — T-1502 proven on at least one real merge. **Not on a green test run**: on a merge that actually landed a matching commit.

This workflow has never run. The tests assert its safety properties from the file with comments stripped, and the guard has an e2e that builds real commits — but nothing here is the real merge that ticket asks for, and shipping it in a release would put an unexercised workflow in the tree that reads as working.

That is #691's shape: 845 lines of installer that no install ran, which three readers in three days took as live.

`workflow_dispatch` does not make it inert either. Whoever dispatches it runs `npm ci` and `build:canonical` on a merged pull request tree, and the second job mints the App key. "Nobody can trigger it by accident" is not the same as "nothing happens when it is triggered".

**What holding costs is only the status quo** — #720 keeps waiting on a maintainer's rebuild — and that cost is already recorded on #719 with countable conditions.

## What it needs before merging

One real dispatch against a real source-only pull request, with the run URL and the resulting canonical pull request linked here. Specifically:

```
the canonical pull request attaches all eleven required contexts
its dist/ matches the merged source
merging it with a merge commit closes the source pull request as merged
the guard skips that merge rather than opening a second rebuild
```

The last one has never been observed. Its shape does not exist in this repository's history — every pull request currently carries its own `dist/`, which is #719's complaint — so the first rebuild merge is the first time the guard meets its subject, and being wrong there is a loop on `main`.

MongLong0214:
## A blind refutation round broke three of the four claims this branch rested on

I put four claims to an independent reviewer with instructions to refute rather than approve, gave it a worktree checked out at this branch's head, and required it to report which files it actually opened. It read 19 files — the 9 it was given plus 10 it went looking for — and broke three.

### C1 — merging this before T-1502's observations exist is safe: **could not refute**

Four attacks failed: this branch does not touch `ci.yml` (its `git diff --exit-code -- dist/ installer/canonical-artifact.json` still runs on every pull request); the workflow is `workflow_dispatch` only so it never fires on an ordinary contributor's pull request; `CONTRIBUTING.md` and `docs/CANONICAL-BUILD.md` still require a local canonical rebuild; and T-1502 deliberately left the committed-`dist/` requirement in place as the fallback. A contributor on today's path is not made worse off by a broken dispatch job.

### C2 — "merge first, then observe" is forced: **refuted**

I had written that merging a source-only pull request would trigger the rebuild. It does not. `canonical-merge.yml` is `workflow_dispatch` with a `pull_request_number` input, and `test/canonical-merge-workflow.test.ts` pins exactly that. Somebody dispatches it; nothing about merging starts it.

### C3 — the plan produces T-1502's two observations: **refuted**

Three separate errors, and the third is the one that matters. A source-only pull request cannot merge through its own button while `ci.yml` still requires `dist/` — `artifact:verify` fails before `git diff` is even reached, which is the trap T-1503 already recorded from #720. Dispatch was missing from the plan entirely. And the negative control the ticket names — skipping `artifact:manifest` — **cannot be performed from a pull request**: that step is hard-coded in a workflow loaded from the default branch, and the source-only filter refuses changes under `.github/workflows/`. A negative control nobody can run is the same defect #722 named.

### C4 — no token leak and no execution of a pull request's head: **half refuted**

The token half survived four attacks: `COMMITLORE_BOT_KEY` appears only in `publish`, that job runs no `npm ci` or `npm run`, the mint script is `git show`n from `main` into a temp directory outside the workspace, and both checkouts set `persist-credentials: false`.

The execute-head half is false, and the file says so itself: rebuilding somebody's change means running it. That is a statement about the ticket, not about this workflow — see below.

---

## Verifying C4 found something none of the four claims was about

`gh pr create` renders this into every canonical pull request it opens:

```
...a merge commit lands that commit on `main` and GitHub closes #123 as merged,
which is what T-1502 asks for.
                  ^^^^^^^^^^^ a closing keyword
```

A pull request closed by keyword is recorded **closed**, `mergedAt` null. So the sentence promising that #N closes as merged is what would deny it — on every run, and there is no API to convert it afterwards. No test read the body.

This was measured six hours ago on #752: #760's body said "GitHub closes #752, #755, #756 … as merged", the keyword bound to #752 alone, and that one was recorded closed while the five with no keyword were recorded merged by reachability.

**Fixed**, and a test now renders the `printf` body and fails on any closing keyword. Restoring the defect makes it fail naming `closes #123`; 21 tests pass across both workflow files with the fix in.

## Two ticket statements were wrong against the file

**"The job never checks out or executes a pull request's head."** Borrowed from the rule #723 fixed for `preserve` — but `preserve` only *reads* a pull request and this one *rebuilds* it. Unsatisfiable by any implementation of this feature, so it would have been quietly dropped rather than met. Amended to the property the `canonicalise` → `publish` split actually holds: the runner executing contributor code carries no credential.

**The negative control.** Replaced with one that can be performed: after the workflow pushes `canonical/pr-N`, add a commit to that branch editing `dist/` without rebuilding, and `ci.yml` must go red. That falsifies the property the ticket claims — the bytes that land match the source that landed with them — rather than the workflow's internal step list.

## One limitation recorded rather than fixed

The canonical pull request asks for a merge commit and cannot enforce one. Squash and rebase are both enabled and GitHub's button remembers the last method used. Recorded in the ticket with both remedies; choosing one is the owner's call, not this ticket's.

Issue #746, opened 2026-08-18
After an upgrade the hook says "cannot find the CLI" when the CLI is fine and containment refused it


MongLong0214:
> **Correction (see [the comment below](https://github.com/MongLong0214/commitlore/issues/746#issuecomment-5322009360)):** the first version of this issue proposed making `commitlore.root` follow `current`. **That would remove a security control** — the pin is a fence, not an oversight. The mechanism below is unchanged; the fix section has been rewritten.

`commitlore hooks install` records two values. One follows `current` by design; the other is pinned to the physical tree that recorded it, also by design. An upgrade moves what the first points at, and the hook compares them — so after every upgrade the recorded pair stops matching, and **the hook reports that as "no CLI found" when the CLI is fine.**

## What is recorded

```
commitlore.bin   <data-root>/current/dist/commitlore.mjs      version-free, so the hook follows
                                                              upgrades (#693)
commitlore.root  <data-root>/v1.0.2                           realpathSync(PACKAGE_ROOT) — the tree
                                                              that recorded the pin, src/commands/hooks.ts:295
```

Both are deliberate, and they are a pair. `src/commands/hooks.ts:284-287` states what `root` is for:

> *"A `.git/config` edit made after this install (ADR-0011's threat model: the same permission that can write this key can write `.git/hooks` directly) can still repoint `commitlore.bin` at another recognized file, **but not at one outside here** — the stub checks the recorded path against this root, not just its name."*

And `versionFreeEntryFor` (`hooks.ts:253-255`) applies the same policy at record time: a `current` that resolves anywhere other than the running bundle is refused, because *"sending a hook to code this one never verified is worse than pinning a version."*

## What the hook then does

`.git/hooks/commit-msg` resolves both with `pwd -P` and requires containment:

```sh
recorded_dir=$(cd "$recorded_parent" && pwd -P)   # follows `current` -> <data-root>/v1.1.2/dist
root_dir=$(cd "$root_slashed" && pwd -P)          # <data-root>/v1.0.2
case "$recorded_dir" in
  "$root_dir"|"$root_dir"/*) exec "$recorded_node" "$recorded" validate ... ;;
```

`pwd -P` resolves symlinks, so `recorded_dir` follows `current` to the new version while `root_dir` stays on the old one. The check fails — **correctly**, since this install has not verified the tree `current` now points at — and the hook falls through to a PATH lookup and a `node_modules` walk.

## Reproduction

[...]

PR #756, opened 2026-08-18
Stop charging every repository for an ordinary upgrade (#749)

### LINKED ISSUE #749: An upgrade leaves every already-wired repository refusing commits, and the repair cannot reach them
Split out of #746, which #748 narrows to the message. This is the part a better message does not fix.

## The operator-visible shape

`commitlore.bin` is recorded through `<data-root>/current` so a hook follows an upgrade; `commitlore.root` is the physical tree that recorded it, so the containment boundary follows nothing. An upgrade moves one and not the other, the trust check refuses, and the hook falls through to a PATH lookup.

In a terminal with `commitlore` on PATH that fallback works and nothing looks wrong. Under the PATH git actually gives a hook — a GUI client, an IDE, a launcher — it is `exit 1`, so **the commit is blocked**. After a single upgrade that is true of every repository wired before it, simultaneously.

Fail-closed is the right direction and is not what this issue disputes. What it asks is whether an ordinary upgrade should land in the same branch as a hostile one.

## Two questions, and the first may come before the second

**1. The repair cannot reach the repositories that need it.**

The stub is written into `.git/hooks` at install time. A repository wired before a stub change keeps the old text until `commitlore hooks install` runs *there* — and `hooks install` is also the command that rewrites `commitlore.root`, i.e. the repair itself. So:

- installing a corrected release fixes nothing on its own
- every affected repository must be visited individually
- `install.sh` cannot do it: it records no list of initialised repositories, and "never touches a repository's `.git`" is a deliberate policy, not an omission

That limitation is worth stating plainly because it is the argument for #742's release notice: **the installer is the only process that sees an upgrade happen, and today it says nothing about the repositories that are about to start failing.** It currently prints `run commitlore init` and is silent about already-wired ones.

**2. Whether the gate should refuse in this specific case.**

`r-failopen` records the current policy: *"with the gate installed and no CLI resolvable, commits are still refused — that is the one hook holding a verdict back"*. That reasoning survives for a genuinely absent validator: an unread message must not be reported as a clean one.

What is worth re-examining is narrower. The refusal happens in an arm shared with #71's planted-`commitlore.bin` case (`test/hooks.test.ts:755-778`), where refusing **is** the whole point — the same permission that can edit `.git/config` can edit `.git/hooks`, and the boundary is what stops a recorded path from executing arbitrary code. An upgrade and that attack are indistinguishable from inside the hook, which is exactly why #748 refuses to name a cause.

So this is not "make it pass". It is: given that the two are indistinguishable, is blocking every commit on every machine after every upgrade the right price for the case that matters? The capture hooks already take the other side of the same question (`exit 0`, *"nothing was checked and found wanting — the checker is absent"*), and the gate is deliberately not them.

Any change here is a fail-closed policy change with its own record, which is why it was kept out of #748: a security decision must not ride through review attached to a sentence edit.

## What #748 does and does not do

Does: names the refusal instead of reporting it as a missing binary, tells the two resolution failures apart, and makes `hooks install` report the root it moved instead of printing `unchanged` after performing the repair.

Does not: change what the gate does, or reach a repository whose stub predates it.

## Related

- #746 — the message, narrowed by #748
- #742 — the release notice; question 1 is a concrete argument for it
- #71 — the containment check and the attack the arm exists for
- #629 — reporting a repair as though it had not happened, twice more in this area

MongLong0214:
Relates to #749 — **it does not close it.** See the last section.

After an upgrade, a repository wired before it refused commits under the PATH git actually gives a hook — a GUI client, an IDE, a launcher — until somebody re-ran `hooks install` there. #746 made that refusal say what it was; this stops it happening.

## I had this backwards, and said so on the issue

I wrote that relaxing the exit code was a fail-closed policy change I should not make alone. **The exit code was never what held the property.**

```
src/hooks/commit-msg.ts
  :260   exec "$recorded_node" ...      ← inside the MATCHING arm only
  :269   commitlore_outside=...         ← the refusing arm sets a variable and falls through
  :291   command -v commitlore          ← then the PATH fallback
```

A path the containment check refuses **is already not executed** by the time anything decides what to print. `test/hooks.test.ts` asserts that directly, with a witness file that must not appear — and it only sees `exit 1` at all because it passes a PATH with no `commitlore` on it.

So the refusal at the end is the ending for *"no CLI could be resolved anywhere"*. Reusing it for *"`current` moved and `root` did not"* was a tax, not a fence.

## An upgrade is distinguishable by shape — and #71 is the opposite shape

```
hooks install   bin  = "<data-root>/current/dist/commitlore.mjs"   literal string
                root = realpath(PACKAGE_ROOT)                       physical v<x>

upgrade         the installer-owned symlink moves to a sibling
                the recorded string is unchanged; the new target sits beside root

#71             the recorded string itself is replaced with an arbitrary .js
                root and node stay as installed
```

A `.git/config` editor can write neither the installer's symlink nor a directory beside its versioned trees. So the arm rebinds the trust root to what `current` resolves to now and **runs the same containment check again**.

**Two weaker rules are recorded as rejected**, in the comment and in a test:

| rejected rule | why |
|---|---|
| share a common ancestor | `/` is a common ancestor, so it admits everything |
| follow `current` wherever `bin` points | satisfied by a planted `/tmp/current/dist/commitlore.mjs` |

The second has its own case, so reintroducing it fails rather than passes.

## Verification

`test/hook-upgrade-rebind.test.ts` builds the `<data-root>` layout rather than describing it, because the property is in shell text only git runs.

| case | result |
|---|---|
| `current` and `root` are the same tree | runs — the control |
| `current` moved to a sibling, `root` left behind | **runs** — this is the fix |
| a planted `.js` outside the install | refused, and the witness file is not created |
| a planted directory imitating the layout | refused |

**Negative control:** removing the rebind fails the upgrade case and leaves the other three passing. 88 tests across the five hook suites pass with it in place.

## Limit

This reaches a repository only when its stub already carries the arm, so one installed before this still needs `hooks install` once — the same boundary #746 recorded, and the reason that command stays the named remedy.

## Why this does not close #749

#749 asked two things. This answers the second — the refusal was a tax rather than a fence, and the tax is gone.

**The first is still true after this ships.** The rebind lives in the stub, and the stub is written into `.git/hooks` at install time. A repository wired before this release keeps the old text, so **the upgrade that delivers this fix is not fixed by it.** Every affected repository still needs `commitlore hooks install` run in it once.

That is question 1 — *"the repair cannot reach them"* — and closing this issue would hide it behind a fix that does not apply to the machines that have the problem today.

## Two corrections to what this PR claimed

**"Does not widen the attack surface" was imprecise, and the precise version is more useful.**

```
somebody who can write .git/config     surface unchanged — measured, and #71's witness still not created
somebody who can write <data-root>     was already trusted, before this change
```

The second was checked on `origin/main`, without the rebind: overwriting `dist/commitlore.mjs` **inside the recorded root** executes the planted code, because that root is exactly what the check trusts. So the rebind does not hand data-root write a capability it lacked — it extends an existing trust to a diverged state.

**A TOCTOU window exists, and it is not this change's.**

```
the check      resolves with `pwd -P` and compares physical paths
the exec       exec "$recorded_node" "$recorded"    ← the unresolved recorded string
                                                      `current` is followed again here
```

Both `exec` sites have it — `:260`, the pre-existing matching arm, and `:299`, the rebind. So it is duplicated rather than introduced, and it is recorded here because it now appears twice and would otherwise be found later as a regression of this PR.


MongLong0214:
## Known residual, recorded rather than discovered later

One arm this does **not** reach, and should not: deleting the old release directory after upgrading.

```
current → v1.1.0, v1.0.0 still present   exit 0   "shape ok · references ok"   ← fixed here
current → v1.1.0, v1.0.0 deleted         exit 1   "One of those two does not exist,
                                                    so the trust check could not run"
```

That is the `commitlore_unresolved` arm, and **the same rule cannot fix it.** The sibling test's anchor is the recorded root; when that root is gone there is nothing to be a sibling *of*, and rebinding without an anchor is a rule anybody can satisfy. Refusing is correct there, and #746 already made the message say which of the two failed.

`install.sh` does not remove old trees — a real data root here holds `v0.8.0` through `v1.1.1` alongside `current` — so reaching that arm takes a deliberate manual delete. This change covers the ordinary upgrade path.

## Where the sibling check actually bears

Worth writing down, because only one axis of the rule is unforgeable:

```
1  pattern match    */current/dist/commitlore.mjs      an attacker can satisfy this
2  [ -L <link> ]    a symlink they created             an attacker can satisfy this
3  parent(root) == parent(realpath(current))           ← this is the one that refuses
```

A `.git/config` editor can build 1 and 2 in `/tmp`. What they cannot do is make their tree a sibling of the root the installer recorded, which is why the planted-layout case is a test rather than a comment.

Forging the whole data root — `bin` and `root` both under one attacker-owned directory — passes, and **passes today as well**: `bin=/tmp/x/v1/dist/…` with `root=/tmp/x/v1` satisfies the existing prefix check unchanged. This does not widen that door.

Issue #221, opened 2026-07-30
test/init.test.ts fails intermittently on node 22 and the cause is unexplained


MongLong0214:
Split out of #192 by owner decision. #192 keeps the diagnostic-honesty defect, which T-1030 closes; this issue keeps the failure itself, which nothing has explained.

## What is observed

`test/init.test.ts` — "formats a human-readable report with a clean summary line when nothing needs attention" — failed once on node 22 and passed on the same commit in a run started four seconds later.

| run | started | node | result |
|---|---|---|---|
| [30492464436](https://github.com/MongLong0214/commitlore/actions/runs/30492464436) | 21:27:51Z | 22 | success |
| [30492468520](https://github.com/MongLong0214/commitlore/actions/runs/30492468520) | 21:27:55Z | 22 | **failure** |

Same tree, same node version, four seconds apart. Node 24 passed in both.

`doctor --fix`, step 4 of 4, reported two failures instead of none:

```
fail    commit-msg hook — ... outcome: the hook fails when git's PATH carries no node:
        file:///home/runner/work/commitlore/commitlore/dist/mcp/server.js:49
fail    hook runtime — the hook fails when git's PATH carries no node:
        file:///home/runner/work/commitlore/commitlore/dist/mcp/server.js:49
```

## What is not known

- Why `dist/mcp/server.js:49` throws. The full stderr is not in the run output, and the probe's message replaced it with an assumed cause — that half is #192's scope.
- Whether node 22 is the trigger or whether its timing merely makes a race more likely. Two data points on one version do not separate those.
- Whether it can affect a real install or only the test's temporary repository.

## Why it is worth an issue rather than a retry

It blocked a release PR's checks until a re-run. A failure that gates a release and cannot be explained is worse than one that fails consistently, because the next occurrence will also look like bad luck.

Reproduction has not been attempted locally on node 22. That is the first step.

MongLong0214:
A **second** test in `test/init.test.ts` has now failed environment-sensitively, which widens this issue beyond the clean-summary case.

Observed on PR #227 at head `0f5ef329d8bccbfa99ca56270794a78344d00b6a`, CI run 30529152230:

```
× commitlore init — a step that cannot fully succeed is reported, not hidden
  > a fresh repo with no remote keeps its warnings and skips visible, but exits cleanly
AssertionError: expected 1 to be +0
Test Files  1 failed | 48 passed (49)
```

- `check (22)` **passed** on the same commit; only `check (24)` failed.
[...]

Issue #192, opened 2026-07-30
init's clean-summary test fails intermittently on node 22, and the probe misreports the cause


MongLong0214:
**Authority binding** — the planning set is merged. Canonical authority is now `dev` at `a333e76f59493ec90e100d1d44c833d7d9e5f093` (PR #220 merged). Exact set: 25 issues #192–#213, #215, #216, #219, one ticket each. Evidence bound to any earlier candidate head is void — re-establish the baseline on this `dev` head before citing a number. Historical evidence SHAs quoted below are left as recorded.

**Ticket**: `T-1030` in `docs/tickets/release.md` (planning candidate: PR #220@a333e76f59493ec90e100d1d44c833d7d9e5f093)
**Milestone**: `M5 · Product loop (Gate A)` · **Gate**: A
**Baseline to preserve**: 45 test files, 1500 passed, 1 skipped; both typechecks exit 0.
**Evidence policy**: evidence is exact-head bound; an unknown cause is reported as unknown rather than guessed.

One responsibility. This ticket corrects diagnostic honesty; it does not claim to fix the intermittent root cause.

---

**Owns**

- `src/commands/doctor.ts` — the hook-failure diagnostic at line 453 (at `dd6dfe2`), specifically the template string `the hook fails when git's PATH carries no node: ${said}`

**Depends on**

- Nothing — this is a standalone honesty fix

**Forbidden scope**

- Do NOT fix the intermittent hook failure itself — this ticket owns the diagnostic message only
- Do NOT modify the release gate, merge anything, or tag anything
- Do NOT touch `src/mcp/server.ts`, `src/hooks/`, or any hook registration logic
- Do NOT alter exit codes or the `check()` function signature
- A root-cause fix for the intermittency is NOT guaranteed by this ticket

**RED test**

- File: `test/doctor.test.ts` (new or addition to existing)
- Reason: a test simulates a hook that exits non-zero with stderr containing a node stack trace (e.g., `at Object.<anonymous> (/path/dist/mcp/server.js:49:1)`) and asserts the diagnostic message does NOT claim "git's PATH carries no node". The test must fail on `dd6dfe2` because the current code unconditionally interpolates `${said}` into the sentence "the hook fails when git's PATH carries no node: ..." regardless of whether the stderr indicates a missing-node problem or a runtime error.

**Minimum GREEN**

1. The diagnostic distinguishes at least two cases:
[...]

PR #757, opened 2026-08-18
Stop a process on the machine from failing a repository command (#750)

### LINKED ISSUE #750: init.test.ts fails on any machine with a stale CommitLore MCP server running
Six tests in `test/init.test.ts` fail locally and pass in CI. The difference is not the code — it is that a CI runner has no other CommitLore processes on it.

## What happens

`init`'s final step runs `doctor`, and the tests assert a clean report:

```
expect(text).toContain('✓ Final check')
expect(exitCode).toBe(0)
```

On a developer machine the report comes back with exactly one warning:

```
! Final check — needs attention
  …
  warn  live MCP runtime identity — 2 distinct live CommitLore runtimes are answering MCP
        <data-root>/v1.0.1/dist/commitlore.mjs (root <data-root>/v1.0.1)
        <data-root>/v1.1.1/dist/commitlore.mjs (root <data-root>/v1.1.1)
  …
init: 1 step(s) need(s) attention — doctor --fix
```

Every other check in the report is `ok`. The failing one inspects **live processes on the machine**, which no temp repository can isolate: an editor session left running against an older install is enough, and older installs are exactly what `<data-root>` keeps around by design.

Failing tests, all for this one reason:

```
the happy path > installs every onboarding component and reports clean codes …
the happy path > is idempotent: a second run changes nothing and still reports 0/0/0
the happy path > formats a human-readable report with a clean summary line …
repository MCP registration > creates a portable registration and clears doctor's …
a step that cannot fully succeed … > a fresh repo with no remote keeps its warnings …
the capture policy step > authorises unattended capture where no policy file exists …
```

## Why it is worth fixing rather than living with

The check is right; the test's assumption is what does not hold. `doctor`'s job is to report machine state, and `live MCP runtime identity` is machine state by definition — that is the whole point of #660 and the reason the check exists.

The cost is that the suite is red for every developer with a running editor session, so a red `init.test.ts` stops carrying information. Twice in one session I attributed this failure to a change of my own before checking `origin/main`, where it fails identically. A test that cries wolf is how a real regression gets waved through.

## Shape of a fix

The tests need `init`'s report to be deterministic, and the honest way is to make the machine-global check not participate in a repository-scoped assertion. Options, roughly in order of preference:

- **Assert on the checks the test actually governs.** `✓ Final check` is a summary over ~20 checks, most of which have nothing to do with `init`. Asserting that the repository-scoped ones passed — hooks, index, notes refspec, MCP registration — says what these tests mean and does not depend on what else is running.
- **Let the harness pin the machine-global inputs**, the way `GIT_CONFIG_GLOBAL=/dev/null` already pins git's. If the live-runtime probe reads a known location, the tests can point it somewhere empty.
- Not: skipping the check under `CI`/`NODE_ENV`, which would make the suite assert a different program from the one that ships.

## Reproduction

```
$ git checkout origin/main
$ npx vitest run test/init.test.ts
  Tests  6 failed | 18 passed (24)
```

with any CommitLore MCP server from a previous install still running. `commitlore doctor` in any repository names them under `live MCP runtime identity`.

MongLong0214:
Closes #750.

Six cases in `test/init.test.ts` failed for every developer with an editor session open, and passed on CI runners. One warning caused it:

```
warn  live MCP runtime identity — 2 distinct live CommitLore runtimes are answering MCP
      <data-root>/v1.0.1/dist/commitlore.mjs
      <data-root>/v1.1.1/dist/commitlore.mjs
```

`<data-root>` keeps old versions by design, so that is the ordinary state of a developer machine rather than a misconfiguration.

## Why it reached the exit code

`init`'s final step is stricter than `doctor` **on purpose** — any check needing attention makes the step incomplete, because `init` is the command that was supposed to take care of everything at once.

That reasoning does not reach a finding about the machine. **The check already made this argument for its own status**, and the comment is still there:

> `warn`, not `fail`. What this observes is the machine, not this repository: a server left running by another session, from an install that has since been deleted, is not something the checkout can fix and must not decide its exit code. **Making it fail also made this suite's result depend on what happened to be running while it ran**, which is the defect class this check exists to surface.

The same sentence applies one level further, and `check()` already has the parameter. The two existing overrides clear attention for the no-remote refspec warn and the ENOENT inject fail on exactly this test — *"neither is something the user can act on here"* (#192, #221).

All three `warn` rows here now clear it. They stay in the report, which is where a machine-scoped observation belongs; they stop deciding whether a repository-scoped command completed.

## The test pins the property, not the symptom

Reproducing the symptom needs a stale server actually running — which is the dependency being removed. A test that needed one would carry the defect it tests for. So the check is driven with an injected scan:

| case | status | needsAttention |
|---|---|---|
| two live runtimes | `warn` | false |
| the scan itself could not run | `warn` | false |
| a runtime whose install was deleted | `warn` | false |
| one runtime answering | `ok` | false |

**Negative control:** removing the override fails three of the four.

## Why this mattered beyond a red suite

Twice in one session I attributed this failure to a change of my own before checking `origin/main`, where it failed identically. A file that is usually red stops carrying information, and that is how a real regression gets waved through.

## Limit

This is one check's classification, not a scope field. Another machine-scoped check added later will default to claiming attention again, and nothing here would notice.


MongLong0214:
One place this is broader than the precedent it cites, recorded rather than left to be noticed.

The two existing overrides each turn attention off for **one branch** of their check:

```
notes-refspec    only remotes.length === 0    — absent, forced and unreachable still claim attention
inject-runtime   only the ENOENT fail          — a missing hook and a non-zero exit still do
this change      every non-ok branch of mcp-runtime-identity
```

So the shape is the same and the reach is not, and it is worth saying why that is not a wider principle arriving by the back door.

**This check has no `fail` path and no repository-keyed branch.** Its three non-ok outcomes are: the process list could not be enumerated, a live runtime is missing `dist/commitlore.mjs` or `spec/SPEC.md`, and two distinct identities are answering. Every one of them is answered by killing a process or reinstalling a cache under `<data-root>` — machine work, none of it reachable from the checkout `init` was setting up. The other two checks have branches that *are* repository work, which is why theirs are per-branch and this one is not.

**And the line is drawn at the observer, not at the subject.** `discoverLiveMcpRuntimes` is the only consumer of `ps` in the checks, and `checkMcpRuntimeIdentity` is its only reader. The sibling checks that mention processes are all keyed to this repository — `mcp-lifecycle` reads `.git/commitlore/mcp-lifecycle.log`, `inject-version` and `inject-runtime` read this repository's hooks, `unattended-initiator` reads this repository's `.mcp.json` — and all keep attention. Exactly one check reads the process table, and it is this one.

**The default is untouched.** `model.ts:246` still computes `needsAttention` from the status, and `grep -rn needsAttention src/commands/doctor/checks/` returns two files: this one and the pre-existing `inject-runtime`. Sharing `warn` does not turn attention off — a check that omits the eighth argument keeps it.

The residual is the one named in the PR's Limit: a machine-scoped check added later defaults to claiming attention again, and nothing here would notice. That is a scope field's job, and this is not one.

PR #763, opened 2026-08-18
Negative control for T-1502 — do not merge

MongLong0214:
**Not for merge.** This is T-1502's negative control, and it is closed once it has gone red.

The control the ticket originally named — skipping `artifact:manifest` — cannot be performed from a pull request. That step is hard-coded in `canonical-merge.yml`, which is loaded from the default branch, and the workflow's source-only filter refuses any change under `.github/workflows/`. A negative control nobody can run is the defect #722 named, so the ticket now names one that can be run: this.

This branch is #762's canonical rebuild with `dist/commitlore.mjs` edited by hand and `src/` left alone. The bundle digest moved `6e69cc37…` → `879e17c8…` with no source change behind it.

CI rebuilds from source and then runs `git diff --exit-code -- dist/ installer/canonical-artifact.json`. **The `check` jobs must fail.** If they pass, *the bytes that land match the source that landed with them* is not a property this repository actually checks, and #762 proves nothing.


MongLong0214:
Went red, in the right place.

```
canonical artifact verified: fcd0832fed28ef33c18591a93d743c918b3af2eda7d311e7bc321d34b5921e8a
diff --git a/dist/commitlore.mjs b/dist/commitlore.mjs
index 1396641..714b9b4 100755
Binary files a/dist/commitlore.mjs and b/dist/commitlore.mjs differ
##[error]Process completed with exit code 1.
```

Both `check (22.23.2)` and `check (24)` failed: https://github.com/MongLong0214/commitlore/actions/runs/32108141891

`artifact:verify` **passed**, and that is the part worth recording. The manifest binds `src/`, and `src/` was not touched — so a hand-edited bundle satisfies the manifest check completely and is caught only by the rebuild-and-diff after it. Without that diff, *the bytes that land match the source that landed with them* would not be a property this repository checks, and #762 would prove nothing.

Closing. T-1502's second acceptance criterion is observed.

PR #765, opened 2026-08-18
F15's ticket file still says unscheduled for two tickets that shipped

MongLong0214:
T-1502 and T-1503 landed today. The file still read *"T-1502 onward remain unscheduled"* — written when ADR-0036 answered how a bot merge is verified without deciding to build one.

A ticket file that says a shipped thing is unscheduled is worse than one that says nothing, because it is the document somebody reads to find out what is left.

- **T-1502** — done, marked against what was observed rather than what was intended: #761 carried source only, `canonical-merge` rebuilt it into #762, and `e4e5154` landed with `artifact:verify` and `git diff -- dist/` both clean, the bundle commit authored by the build App. #763 is the control that went red.
- **T-1503** — done, `532c30f4`.
- **T-1504** — still unscheduled, and the header now says why that is not a scheduling decision. Its remaining condition is an observation about a person — a non-Linux contributor completing a source change unaided — and nothing in this repository can produce one.

Documentation only; the file owns no assertion. 23 tests across the two canonical suites pass unchanged.

### DIFF
--- a/docs/tickets/F15-canonical-artifact-provenance.md
+++ b/docs/tickets/F15-canonical-artifact-provenance.md
@@ -10,8 +10,18 @@ The App does not push to `main`; it opens a pull request. `main`'s protection re
 `lint`, which runs only on `pull_request`, so no push can satisfy protection on its merits
 and every push-shaped option needs a bypass. Going through a pull request needs none.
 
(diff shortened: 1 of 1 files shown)


MongLong0214:
Superseded by the revert of T-1503. The ticket-status edits here said T-1503 was done; it is now rejected, and the corrected statuses ship with the revert.

Issue #776, opened 2026-08-18
A fresh clone serves empty answers until the index finishes, and which records are missing is whichever the budget did not reach


MongLong0214:
In a freshly cloned or materialised repository, the first answers are empty because the index build stops at a three-second budget — and whether a record arrives depends on which commits it happened to reach.

## Observed

A worktree materialised from a bundle of this repository, then the shipping hook command run against a path that has records:

```
$ inject --hook-input < {"tool_name":"Edit","tool_input":{"file_path":"src/commands/pending.ts"}}
stdout: 0 bytes
stderr: commitlore: the index is incomplete: the build stopped after its time budget
        with 588 commit(s) or note(s) unread — records in them are missing from this answer.
```

Completing the index and repeating the identical command in the identical worktree:

```
$ index --rebuild      rebuilt: scanned 652 commits, 4113 trailers in 23075ms
$ inject --hook-input  stdout: 3201 bytes, contains r-gcunstageable
```

The only variable was whether the index had finished. Twenty-three seconds of work, against a consumer budget of three (`CONSUMER_SCAN_BUDGET_MS`).

## Why this is worse than "the first answer is slow"

The budget is not a hard cutoff on *time to answer* — it is a cutoff on *how much history the answer is derived from*, and which part gets read is whichever the scan reached first. So:

- an agent editing a file whose record sits in a recently-read commit gets the record;
- an agent editing a file whose record sits past the cutoff gets an empty payload;
- the same repository, the same command, a different path, and no way to predict which.

The product is careful everywhere else to distinguish *no record exists* from *this scan did not reach it* — `coverage: partial`, `history: unavailable`, `notes: unfetched` are separate states for exactly this reason. That distinction is preserved here: the stderr line names the unread count. **The honesty layer works. What is missing is that nothing finishes the index before the first answer is served**, so the disclosure fires on ordinary first use rather than on an exceptional one.

## And in this repository the remedy fails

The stderr line says `fix: commitlore init (or commitlore index)`. In a worktree whose HEAD is behind the notes ref, `index --rebuild` exits 2 with `NOT NULL constraint failed: trailers.signature_status` (#774). So the advice is correct and, in that state, not followable.

The two together mean a fresh clone can be left permanently serving partial answers with no path out that the message names.

## Where this bit

[...]

Issue #775, opened 2026-08-18
The plugin and the CLI install different hook matchers, and the docs describe only one


MongLong0214:
The plugin and the CLI install different hook matchers, and neither covers every tool `inject` can answer for.

```
hooks/hooks.json          "Edit|Write|MultiEdit|NotebookEdit"     no Read
claude-settings.ts:33     "Read|Edit|Write"                       no MultiEdit, no NotebookEdit
inject.ts PATH_TOOLS      Read, Edit, Write, MultiEdit, NotebookEdit
```

`inject --hook-input` extracts a path from all five. Which of them ever reaches it depends on how the user installed:

| install path | fires on | silent on |
|---|---|---|
| plugin (`/plugin install`) | Edit, Write, MultiEdit, NotebookEdit | **Read** |
| `commitlore hooks install` | Read, Edit, Write | **MultiEdit, NotebookEdit** |

So a repository set up through the CLI delivers nothing when the agent edits with `MultiEdit`, and one set up through the plugin delivers nothing when the agent reads a file before deciding. Same product, same command behind the hook, different coverage.

## How it surfaced

A benchmark cell that delivered records under the older matcher delivered none under the current one:

```
pilot (matcher Edit|Write|MultiEdit|NotebookEdit)   hook_invocations 4 and 8, records delivered
rerun (matcher Read|Edit|Write)                     hook_invocations 0, records delivered 0
```

Same task, same snapshot, same agent, opposite delivery. Injection itself is fine — running `inject --hook-input` by hand on that path in a fresh clone returns a 3.2 KB payload. The hook simply never ran, because the tool the agent used is not in the matcher the CLI writes.

## The documentation describes only one of them

`docs/COMPATIBILITY.md:215` states the hook as `PreToolUse` on `Edit|Write|MultiEdit|NotebookEdit`. That is the plugin's matcher. `ADR-0026` and `docs/tickets/F14-distribution.md` say the same. Nothing documents that `hooks install` writes a different set — so a reader checking whether their setup covers `MultiEdit` is told yes, and for the CLI path the answer is no.

## What the fix has to decide

Not simply "use the union". The two matchers differ in a way that looks deliberate on each side:

- **`Read` is in the CLI set and not the plugin's.** Firing on every `Read` is what makes the shipping surface heavier than the plugin's, and that cost is real — the benchmark work measured a lower bound of 1.45x on token overhead partly from it.
[...]

Issue #774, opened 2026-08-18
index --rebuild cannot rebuild past a schema change, and dies with a SQLite message


MongLong0214:
`commitlore index --rebuild` exits 2 with a SQLite constraint error when HEAD is behind the notes ref.

```
$ commitlore index --rebuild
commitlore: NOT NULL constraint failed: trailers.signature_status
$ echo $?
2
```

## Reproduction

```sh
git worktree add --detach /tmp/old <a-commit-a-few-hundred-behind>
cd /tmp/old
commitlore index --rebuild
```

At the repository's current HEAD the same command succeeds — 1091 commits, 7083 trailers, ~47 s. Detached a few hundred commits back, it fails every time.

## Where the null comes from

`trailers.signature_status` is declared `TEXT NOT NULL` (`src/core/index-db.ts:223`) and caches Git's `%G?` for the commit the trailer was read from (`:1089`).

It is not the commits. In the snapshot's own history every commit has a value:

```
224  E
428  N
  0  (empty)
```

It is the **notes**. `refs/notes/commitlore` is a repository-global ref and is not filtered by the scanned commit range. In this repository all eleven notes are attached to commits that are *not* ancestors of the older HEAD:

```
notes total                                     11
attached to a commit outside the scanned range  11
```

So the rebuild reads a note, has no commit in its set to take `%G?` from, and writes null into a NOT NULL column.

## Why it matters beyond an old checkout

Anyone bisecting, or working in a worktree pinned to a release tag, hits this — and the failure names a SQLite column rather than the situation, which is the shape #746 fixed elsewhere. The message does not say that the index was rebuilt from a HEAD behind the notes ref, and does not suggest what to do.

A benchmark replay is how this was found: the snapshot's index could not be completed, so every answer from it carried `coverage: partial` and the conclusions drawn from *absence* were unsafe. The path-scope conclusions in that work were re-derived from `git show --name-only` instead, which is index-independent, and agreed — but they had to be re-derived.

## What a fix has to decide

Whether a note whose commit is outside the scanned range should be:

- **skipped**, with the count surfaced the way unread commits already are (`the index is incomplete: … N commit(s) or note(s) unread`); or
- **indexed** with a signature status that says *not evaluated here* rather than borrowing one, since `%G?` is the verifying process's verdict about a commit this scan never read.

The second is the honest one if those trailers are meant to be answerable at all. Either way the column stops taking a null, and the failure stops being a constraint name.

[...]

PR #777, opened 2026-08-18
The ON arm was not receiving the treatment, and the harness could not tell

MongLong0214:
Seven commits. Two make the harness administer the treatment and record what happened; five are assertions about what the benchmark currently cannot measure, committed **failing**, so the work they specify cannot be forgotten.

## The harness was not running the study it describes

**The ON arm received nothing.** A materialized worktree has no index, so the first `inject` builds one under the three-second consumer budget and stops partway.

```
before   hook_invocations 0, delivered 0    inject by hand: 0 bytes, "588 commit(s) unread"
after    hook_invocations 8, delivered 4    what the pilot's second repeat recorded
```

`hook_invocations` counts a payload header, so an empty payload reads as a hook that never fired. Built in **both** arms — only in ON and the arms differ in setup as well as treatment.

**The session was thrown away.** `--output-format json` means stdout is the whole session; it was parsed for `usage` and dropped. Enabled for one run, it overturned the conclusion of eight:

> **Why not a blanket `--force`:** this repo's own design-record log (surfaced automatically when I read `pending.ts`) already shows a plain `--force` covering "staged or unreadable" was proposed and ruled out …
>
> **What I built instead:** `pending rm --force` now peeks the phase before touching the file. If it reads as `staged`/`applied`, **it still refuses unconditionally, even with `--force`.**

Off by default: a session holds prompts, file contents and model output from the studied repository.

## What the benchmark cannot currently measure

Each of these is an `it.fails` — the assertion states what has to be true, the runner records that it is not, and the day it holds the line has to be flipped to `it`.

**An untouched tree scores a decision-safe success on all four tasks.** §4.7's good control asserts `functional_pass === true` on the repository as it stands, which forces the field to mean *the tree is not broken*. Every `functional_pass` is then a substring the target cannot lack — `"guard"` in `guard.ts`, `"rm"` in `pending.ts` — so `decision_safe_success` collapses to `!revived` and doing nothing passes.

**Three of the four revival checks misread correct designs.**

| task | what it rules out | verdict |
|---|---|---|
| `lifecycle-fourth-value` | a fourth member in a union | **sound** — parses the declaration and counts |
| `pending-rm-force` | a force escape **on staged or unreadable files** | not decidable statically — a design that honours it contains the same token |
| `verify-scope` | a declared list of result files | false-positives on `[".jsonl", ".json"]` and on a **comment** |
| `guard-blocking-policy` | a policy key turning guard blocking | false-positives on an unrelated `strict: boolean`, a different schema, and a **comment** |

The comments matter most: this repository's practice is recording why an approach was rejected, so a comment saying exactly that is the likeliest thing an honest implementation contains — and two checks read prose as code.

`pending-rm-force`'s fixture is the real implementation from the run that produced it, kept verbatim. No substring search can pass it.

## Nothing tested the harness

Both fixes above vanish without a symptom — an untreated arm still emits a valid row. Five assertions read the source with comments stripped, pinning order as well as presence.

They worked immediately: `git checkout origin/main -- run.ts`, used to check the tests fail without the change, **stages** what it checks out, so the commit adding those tests also removed the property they pin. CI failed on exactly those five.

## Verified

```
index fix        0 → 8 invocations, 0 → 4 records, same task and snapshot
session          5044 bytes naming the ruled-out alternative and its reasoning
harness tests    restoring main's run.ts fails all five
none control     inverting the assertion so the body passes fails all four
oracle cases     all five fail today; inverting them makes the it.fails cases fail
                 and neither oracle fires on the untouched base tree
```

Findings on #771. Product defects this surfaced: #774, #775, #776.

### DIFF
--- a/bench/cdeb/pilot/run.ts
+++ b/bench/cdeb/pilot/run.ts
@@ -178,6 +178,30 @@ const runOne = (
   gitOrThrow(workdir, ["config", "user.name", "operator"]);
   gitOrThrow(workdir, ["config", "--add", "commitlore.trustedAuthor", owner]);
 
+  // The index has to be finished before the agent starts, in BOTH arms.
+  //
+  // A materialized worktree has no index, so the first `inject` builds one under
+  // the three-second consumer budget and stops partway. Which records that
+  // budget reaches depends on where the scan got to, so an ON run could deliver
+  // four records or none for the same task -- measured here: the same worktree
+  // and the same command returned 0 bytes with `588 commit(s) unread`, and 3201
+  // bytes containing the expected record after `index --rebuild` finished
+  // (23s). An arm that receives nothing is not the arm the study names.
+  //
+  // Run in both arms so the two differ only in whether a hook reads the index.
+  // Its cost is outside the agent session either way, and leaving it out of OFF
+  // would make the arms differ in setup as well as in treatment.
+  const indexed = spawnSync("node", [CLI_ENTRY, "index", "--rebuild"], {
+    cwd: workdir,
+    encoding: "utf8",
+    timeout: 5 * 60 * 1000,
+  });
+  if (indexed.status !== 0) {
+    throw new Error(
+      `index --rebuild failed in ${workdir} (status ${String(indexed.status)}): ${indexed.stderr ?? ""}`,
+    );
+  }
+
   const exposureLog = join(scratch, "exposure.log");
   const settingsPath = armSettings(scratch, condition, exposureLog);
   const mcpPath = emptyMcpConfig(scratch);
--- a/bench/cdeb/pilot/run.ts
+++ b/bench/cdeb/pilot/run.ts
@@ -165,9 +165,7 @@ const runOne = (
   identity: ReturnType<typeof createRepositoryBundle>,
   studyId: string,
   offIdentityByTask: Map<string, MaterializedIdentity>,
-  sessionDir: string | null,
 ): PilotRow => {
-  const logicalRunId = `${task.task_id}__${condition}__r${String(repeat)}`;
   const scratch = mkdtempSync(join(tmpdir(), `cdeb-p-${task.task_id}-`));
   const workdir = join(scratch, "wt");
   const materialized = materializeBundle(identity, bundlePath, workdir);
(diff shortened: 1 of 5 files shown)


MongLong0214:
Second commit on this branch: **keep the agent session**, off by default behind `--sessions <dir>`.

It was written for one run and overturned the conclusion of eight.

The agent's own summary, from a session this harness previously discarded:

> **Why not a blanket `--force`:** this repo's own design-record log (surfaced automatically when I read `pending.ts`) already shows a plain `--force` covering "staged or unreadable" was proposed and ruled out …
>
> **What I built instead:** `pending rm --force` now peeks the phase before touching the file. If it reads as `staged`/`applied`, **it still refuses unconditionally, even with `--force`.**

It read the record, cited it, understood the reasoning, and built something that honours the ruling — with tests for the refusal.

The oracle scored it `rejected_decision_revived: true`:

```js
const revived = /--force|'-f'|"-f"|\bforce\b\s*[?:]/.test(source);
```

A grep for `--force` cannot tell the rejected approach from a design that respects it. Seven variance runs and both pilot repeats carry that label; not one has been shown to have revived anything. Retracted on #771.

**None of that was in the row.** Delivery counts and a binary outcome, and nothing between them — which is why the ambiguity a `[claim]` payload creates was unanswerable from the recorded fields.

Off by default because a session holds prompts, file contents and model output from the studied repository, and that belongs in the authorization first. Written before the JSON is parsed, so a session that fails to parse is still on disk.

MongLong0214:
## Correction on what the persisted session contains

I described this as *"the whole session"* and said it would make derivable *"whether the delivered Record-Id was mentioned, whether the ruled-out alternative was named and then abandoned, how far into the session the first edit happened"*. Reading one of the files it produced: only the first of those is answerable.

`--output-format json` emits the run's **summary**, not its transcript:

```
keys      is_error, num_turns, stop_reason, session_id, usage, modelUsage,
          total_cost_usd, permission_denials, result, uuid, timings
num_turns 41
result    1725 characters — the agent's final message
```

Forty-one turns happened and none of them is in the file. What is there is the final message, and that was enough to overturn the oracle conclusion — the agent explained in it that it had read the record and why it built something narrower. That was luck as much as design: an agent that had done the same work and summarised it as *"added `pending rm --force`"* would have left no trace of having read anything.

So the claim in this pull request should be read narrowly:

```
answerable      did the final message reference the record or the alternative
not answerable  which tools ran, in what order, what was read before the edit,
                whether the record was opened at all
```

The second list is what separates *read and not followed* from *never read* in the general case, and it needs `--output-format stream-json`, which records each turn.

This also means the session cannot answer #775's question — whether an agent edits through `MultiEdit`, which the CLI's matcher does not cover. I went looking for that in these files and it is not there.

**Not changed here.** Switching output formats changes what the harness parses for `usage` and what a run costs to store, and it should be one deliberate change rather than a correction bolted onto this one. Recorded so the next person does not discover the gap by looking for tool calls that were never written down.

PR #780, opened 2026-08-18
Three places named the tool set and all three disagreed (#775)

### LINKED ISSUE #775: The plugin and the CLI install different hook matchers, and the docs describe only one

MongLong0214:
Closes #775.

Three places named the set of tools worth injecting for, and all three disagreed.

| Site | Set | What it silently dropped |
|---|---|---|
| `src/hooks/claude-settings.ts` (CLI install) | `Read\|Edit\|Write` | `MultiEdit`, `NotebookEdit` |
| `hooks/hooks.json` (plugin) | `Edit\|Write\|MultiEdit\|NotebookEdit` | `Read` |
| `src/commands/inject.ts` | all five | — |

Each install had a hole the other did not, and **neither hole was visible from the file that had it** — you had to read the other installer to see it.

### The hole is real

If the matcher were a regex, `Edit` would already cover `MultiEdit` by substring and the CLI would have had no gap. It is not: a matcher of only letters, digits and `|` is evaluated as a list of *exact* strings. `Edit` does not cover `MultiEdit`. Spelling every name out is the mechanism, not verbosity — that is now recorded in `path-tools.ts` where the next person would otherwise assume it.

### Why `Read` stays

`Read` is the expensive member, so it decided the question.

- **ADR-0006 specifies this hook** as firing *"when an agent reads or modifies a file"*. PRD-F4's story is *"when I open a file"*. Dropping `Read` would contradict a decision record, not implement one.
- **SPEC §5's** *"surfaced before an agent edits the path it constrains"* is a requirement on **timing**, and a read satisfies it earlier than an edit does. It is not a narrower surface.
- It is where the record arrives **before** the agent has committed to an approach. Firing only on edits delivers a correction, not context.

`MultiEdit` and `NotebookEdit` are close to free — they *replace* an `Edit` call rather than adding one, so including them does not raise the fire count.

### Measured, not inferred

Run against the built CLI:

| Path | Payload | Wall time |
|---|---|---|
| `LICENSE` (no records) | **0 bytes** | 326 ms |
| `.gitignore` (has records) | 3134 bytes | 376 ms |
| `package.json` (has records) | 3282 bytes | ~530 ms |

A record-free path costs no tokens — but it still costs **~330 ms**. That latency is node start plus opening the database, so it argues for making the fire cheap, not for delivering less. Filed separately.

### One name, and a test where derivation is impossible

`src/core/path-tools.ts` owns the set; both writers derive from it. It is a leaf with no imports because `commands/inject.ts` runs on every fire.

`hooks.json` is JSON and cannot import, so the site that **cannot** derive is the site that gets a test. `test/hook-matcher-parity.test.ts` earned its place immediately by catching a stale row in `COMPATIBILITY.md`.

`test/inject.test.ts` keeps a literal assertion of the shipped string on purpose — a test that derives from the constant cannot catch a wrong constant.

### Known limit, widened by this change

Neither installer knows the other exists. `init` writes the settings.json hook unconditionally, so a user who follows the README to the plugin and then runs `commitlore init` carries **two** PreToolUse hooks running the same command. Unifying does not create that double fire, but it widens the overlap from `Edit`/`Write` to all five. The partial overlap it replaces was worse to diagnose — the same user saw records twice on an edit and once on a read. Filed separately.

### Preregistered numbers left alone

`bench/cdeb/PRD.md` named the old shipping matcher. It gets a dated correction rather than a rewrite: the ⚠ still stands, because the pilot surface remains lighter than shipping. Only *how much* lighter changed, not the direction. The anchors are not re-derived.


MongLong0214:
Closed in favour of #783, which is the same change.

**What was wrong with this one:** both commits carried CommitLore trailers but no `Record-Id`. `test/dogfood.test.ts` rejects that, correctly — a record without an identity cannot be superseded, chained, or expired, so it is inert (SPEC §3.2). The repository applying its own protocol to itself is exactly what that test is for, and it caught me.

I did not amend and force-push over this branch; #783 rebuilds the work as one properly-recorded commit on a fresh branch instead.

Everything else on this branch survives unchanged, plus three fixes that came out of an adversarial review after it was opened: the `F14` capability table (which claimed to be asserted and was not), a dated note on `ADR-0026`, and the parity test extended to cover both documents.

PR #787, opened 2026-08-18
Release awareness: the whole of F16 (closes #742)

### LINKED ISSUE #742: No way to learn a newer release exists — a stale install quietly enforces an older protocol

MongLong0214:
Closes #742. **All seven F16 tickets** — the two tickets with no output, which is why they come first: they can be tested exhaustively before anything prints or reaches a network.

Not closing #742: T-1603 through T-1607 (`upgrade --check`, the passive notice, doctor's finding, `upgrade` itself, and `init --upgrade`) are still open.

### T-1601 — rank release tags the way the installer already does

`install.sh` picks the newest tag without `sort -V` (a GNU/BSD extension it cannot use, being POSIX sh) by zero-padding the three fields so a lexical sort becomes numeric, and by considering only `vMAJOR.MINOR.PATCH`.

**Two rankings that disagree is a defect with no symptom until the day it has one**, and the day is `v9` against `v10`. So the central test does not restate the rule — it runs the installer's own pipeline over the same tag list and asserts both pick the same tag:

```sh
grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | awk -F. '{ printf "%010d %010d %010d %s\n", … }' | sort | tail -1
```

Nothing throws. `runtimeIdentity` reports `0.0.0-unknown` when the manifest is unreadable, and *"we do not know what is running"* must not become *"you are out of date"*. A version that does not parse answers **not newer** — the direction `gh` also takes.

### T-1602 — ask git, once a day, never fatally

**`git ls-remote --tags --refs`, not the GitHub API.** PRD-F16 records the correction: revision 1 justified the API by claiming the install path already used it, and `install.sh` contains no `api.github.com` call at all.

**A failed check is cached, by kind.** `gh` writes state only after a success, so an offline machine retries on every invocation forever — an outbound attempt per command, which is what an egress-monitoring org flags. But one blanket interval is the opposite error:

| Outcome | Waits | Why |
|---|---|---|
| unreachable | 1 h, doubling to a 24 h ceiling | a five-minute outage must not buy a day of silence |
| refused | 24 h | the remote answered; the question is settled |
| no tag matched | 1 h, logged under `--debug` | a bug in this code, not weather — burying it hides it |

**Cancellation is a mechanism, and each part carries weight.** Own process group (so the signal reaches the SSH client or credential helper `git` spawned, not only `git`); a hard timeout independent of the caller (`commitlore --version` returns in ms — cancel-on-completion bounds nothing); `SIGTERM` then `SIGKILL`, because a remote that accepts and never answers will not leave politely; `GIT_TERMINAL_PROMPT=0` so it cannot block on input nobody will type.

**The two switches are not the same switch**, and a test asserts it. `COMMITLORE_NO_UPDATE_CHECK` stops the check; `COMMITLORE_NO_AUTO_UPDATE` stops `init`/`upgrade` from *acting* and leaves reporting alone — someone who declines automatic action has not asked to stop being told. `DO_NOT_TRACK` and `NO_UPDATE_NOTIFIER` honoured.

Context gates (CI, TTY, hook subcommand) are deliberately **absent** — they belong to T-1604. A module that applied them is one `upgrade --check` could not use, because that command answers inside CI on purpose.

### How this is tested

Every test runs against a **local bare repository** through `COMMITLORE_INSTALL_SOURCE` — real tags, real `git ls-remote`, no network. The timeout is driven by an **injected timer**, because a property timed against a real clock passes vacuously on a slow machine.

Negative controls, both confirmed to fail the right tests and only those:

- switching the major-field comparison to string ordering → fails the `v10`/`v9` case and the install.sh cross-check
- collapsing the two switches into one, and giving every failure the same interval → fails exactly the two tests that separate them

One thing worth recording: the first version of the timer double re-fired on every `setTimer` call, so the timeout callback scheduled a grace timer that re-entered the timeout callback, and the suite spun until it was killed. **A test double that calls back into the thing it doubles needs a stop.**

### Names, taken from the tickets rather than from memory

The verb is `upgrade`; `--apply` is gone. But the ticket owns `src/commands/update.ts` — the file name and the command name deliberately differ, and both are as specified (F16 lines 94 and 183). Nothing in this PR registers a command yet.


MongLong0214:
**T-1603 added** — `commitlore upgrade --check` / `--json`, the read-only form. This PR now carries T-1601, T-1602 and T-1603.

Two assertions in it are the ticket rather than decoration:

- **Nothing but the check's own `git ls-remote` is started.** `git` is replaced with a recorder for the length of a real run, and every call it logs is asserted to be `ls-remote`. ADR-0037 says the CLI does not replace itself; a comment saying so is not a guard (#723).
- **It answers inside CI and off a terminal.** The notice owns the context gates and this command must not share them, or the one scriptable form would be silent in the one place scripts run.

*"We could not look"* and *"you are up to date"* are different answers, so `updateAvailable: false` is not allowed to carry both — an `unknown` field says which happened, tested for unreachable, refused, and switched-off-by-name.

The install line is read out of the README with the target tag substituted, and the Windows path prints `install.ps1`. Naming a command that cannot work is the failure `gh` sidesteps by printing a URL.

### Two defects this turned up in my own earlier work

- `latestRelease` took its switches from the env it was handed and its cache location from `process`. Two tests passed by reading each other's answer out of the developer's real cache.
- The first CI run on this branch failed on **canonical dist**: I committed source without the docker build and manifest in the same commit. `dist/` is tracked and the manifest pins the source hash, so a later dist-only commit would still leave the first red. Both are in this push.

### Still open on #742

T-1604 (the passive notice and its context gates), T-1605 (doctor's finding), T-1606 (`upgrade` performs the upgrade — the four-step installer call and link read-back), T-1607 (`init` says what it pinned; `--upgrade` moves it).

MongLong0214:
**All seven F16 tickets are in.** This PR now closes #742.

| Ticket | What landed |
|---|---|
| T-1601 | ranking, cross-checked against `install.sh`'s own pipeline |
| T-1602 | `git ls-remote`, once a day, failure cached **by kind** |
| T-1603 | `upgrade --check` / `--json`, read-only and enforced |
| T-1604 | the passive notice, zero-latency by construction |
| T-1605 | `doctor` carries staleness — the only place `--json` can |
| T-1606 | the four-step upgrade, with a real #735 fixture |
| T-1607 | `init` names what it pinned; `--upgrade` moves it |

### The two assertions the tickets were rewritten to force

**The #735 fixture builds the machine, not the outcome.** A stub that exits 0 without touching `current` catches a CLI that trusts an exit code — worth catching, and not the claim being defended. From `install.sh:603-671` the wrapper is renamed *before* the `current` move, so the real state is a wrapper on the new checkout, a `current` still on the old one, and a stray `current.commitlore-install.<pid>` beside it. The fixture builds all three and each is asserted, so a future fixture that drifts back to a stub fails visibly.

**Step 2 asks whether the link is right, not whether it changed.** Reverting to revision 1's weaker comparison fails **five** tests, including a fixture that moves `current` to a *different* installed version — exits 0, changes the link, still wrong.

**`init` never moves `current`**, asserted with every gate open by shadowing `sh` and `git` on `PATH` and recording every process a real run starts.

### A defect the work surfaced in itself

`--upgrade`'s three failure modes are three facts, and the distinction was broken one level down: a **DNS failure was landing as `refused`**, because `git` exits non-zero for both and only stderr says which. Refused waits a day and unreachable waits an hour — so the misclassification bought a day of silence about exactly the staleness this feature exists to expose. Fixed, with both directions pinned.

Two more caught before CI: `latestRelease` read its switches from the env it was handed and its cache path from `process`, so two tests passed by reading each other's answer out of the real cache; and the doctor snapshots now pin `COMMITLORE_NO_UPDATE_CHECK`, because a row that asks a remote reports one thing on a connected laptop and another on an air-gapped runner.

### Negative controls, every ticket

Each was confirmed to fail the right tests and only those — string-ordered majors, collapsed switches, one blanket cache interval, the platform install line, the `--json` evidence field, revision 1's step 2, and the classification above. One caught a **vacuous test**: `finishUpdateCheck` asserted silence on a failed command while nothing had resolved, so it passed through the wrong branch and kept passing with the gate deleted.

Issue #653, opened 2026-08-14
the index caches a signature status that is not a property of the repository, so two routes grade the same record differently


MongLong0214:
F-003's field report — the same repository, path, record and moment returning `directive` on one route and `claim` on another — reproduces from current `main` with a single build. Two runtimes are not required.

## Reproduced

A repository in signature mode, with the signing key in the allowlist:

```
commitlore.requireSignedDirective=true
commitlore.trustedAuthor=Router <router@test.invalid>
commitlore.trustedSigner=F64FCAD1D09B89E137C25C6BA5ED7A1AF64EE2EF
```

`git log -1 --format='%G?'` reports `G` and `%GF` reports that same fingerprint.

```
1) index built while the keyring was unreachable   => claim
2) keyring restored, same index                    => claim
3) same moment, --no-index                         => directive
```

**Steps 2 and 3 are the same repository, the same record, the same moment and the same binary.** They disagree.

Deleting the index and rebuilding it with the keyring reachable gives `directive` on every subsequent run, which confirms the cached value is what decides.

## The mechanism

`src/core/index-db.ts` stores `signatureStatus` from `%G?` alongside the record (`:699`, `:715`, `:847`, `:862`), and `src/core/grade.ts:806` refuses `directive` when it is not `G`.

But `%G?` is **not a property of the repository**. It is the verdict of the process that ran `git log`, and it depends on which keys that process could reach — the keyring, `GNUPGHOME`, whether `gpg` is installed, whether the key has since been imported. The same commit yields `G` in one environment and something else in another, with no change to the repository at all.

ADR-0003 states the index is a derived, discardable cache of what git already holds. A cached `%G?` breaks that: it is not derivable from the repository, and nothing invalidates it when the environment that produced it changes. The record's authority then depends on the conditions under which the cache happened to be warmed.

## Why this matters for #631 and F-001

The field report was attributed to a stale runtime, and #644 makes surfaces converge on one runtime identity. **That is necessary and not sufficient.** This divergence needs only one build: any surface reading through the index disagrees with any surface that bypasses it, whenever the cache was warmed without the keys. Closing #631 on runtime convergence alone would leave the reproduction above intact.

## Suggested acceptance

- a signature status is not served from the cache as if it were a fact about the repository — either it is recomputed where it is used, or the cached entry records the verifying environment and is invalidated when that changes
- a fixture that warms the index without the keys and then reads with them present grades identically on the index and `--no-index` routes
[...]

PR #789, opened 2026-08-19
A rebuild started one git process per trailer paragraph (#776)

### LINKED ISSUE #776: A fresh clone serves empty answers until the index finishes, and which records are missing is whichever the budget did not reach

MongLong0214:
**Blocked only on CI. Nothing else is outstanding — all three items from the #776 audit are in, measured, and negative-controlled. Merge when green; that closes #776.**

Closes #776. **The 23-second rebuild was not a floor — it was 4,329 git process spawns.**

Not closing #776: this removes the reason the number was large, and the remaining product question (below) is unchanged in shape but much smaller.

### What one rebuild actually did

```
4329  git interpret-trailers --parse --no-divider     ← essentially all of the time
   3  git log --name-only                             (paths)
   2  git log %(trailers:only,unfold)                 (the batch reader)
   2  git log %B                                      (the re-read)
   ...
4345  total
```

The batch reader already avoids one process per commit, and the module says so in its own header:

> *"One `git log` per batch of commits replaces one process per commit, which is the difference between indexing 100k commits and not finishing."*

`explodeRecordBlocks` went around it. `%(trailers)` returns only a message's **last** paragraph, so recovering a squashed earlier block means re-reading the message — and that pass parsed *every* earlier paragraph to find the ones carrying a second record. **On this repository: 3,900 parses to keep 33.**

### Both changes move a cheap test in front of an expensive one already being discarded

`parseRecordBlocks` ran `asIsolatedBlock` on each earlier paragraph and **then** dropped any result without a `Record-Id`. The raw-text test now runs first — a trailer key is always at the start of a line in the source and folding continues *values*, so a paragraph that never names the key cannot yield one.

`explodeRecordBlocks` skips the parse entirely for messages the atom pass already covers.

### The gate was wrong the first time, and how is worth reading

ADR-0014 says *"a message with at most one `Record-Id` anywhere has exactly one block"*, so the first gate counted occurrences and skipped at ≤1.

That is a statement about **how many record blocks exist**, not about **whether the one git handed the atom pass is it.** A commit whose final paragraph is `Co-authored-by:` has its record one paragraph earlier — `%(trailers)` returns the conventional block, and the real record is only reachable through the explode pass. The count-only gate dropped **nine rows** from exactly such a commit here.

The gate now checks position as well as count. The boundary is four tests.

### Measured

| | Before | After |
|---|---|---|
| rebuild, 1122 commits | **45 s** | **6.8 s** |
| `trailers` rows | 7250 | **7250** |

**Identical, not merely smaller** — every row of `trailers` dumped before and after and compared field by field. That check is what caught the first gate; the timing alone looked like a win.

Negative control: reverting to the count-only gate reproduces the nine missing rows.

### What is left on #776

The product question does not go away, it shrinks: **what first-use delay is acceptable in a freshly cloned repository**, now against ~7 s rather than ~23 s, with `CONSUMER_SCAN_BUDGET_MS = 3_000` still the consumer budget.

Two further reductions are identified and **not** taken here:
- `%G?` in the rebuild's `git log` format costs **2.7 s** on this tree and is only needed in signature mode.
- The 3-second consumer budget still spends itself on explode for the first slice of commits, which is the wrong thing to spend it on.


MongLong0214:
**Second optimisation added: `%G?` off outside signature mode.**

| | rebuild, 1122 commits |
|---|---|
| before either change | **45 s** |
| after the process-spawn fix | 6.8 s |
| after this too | **2.5 s** |

`%G?` makes git verify every commit's signature. Nothing reads the answer unless signature mode is on — `grade.ts` consults `signatureStatus` only behind `requireSignedDirective`, and `trusted-authors.ts` says as much where the verifier generation is defined:

> *"Only signature mode pays for it: the setting is opt-in."*

The rebuild paid for it always.

**Turning the setting on later is already handled**, and by the mechanism built for it: `signatureVerifierGeneration` goes `null` → hash, `healthProblem` compares it against `meta`, the mismatch rebuilds. Same path a changed keyring takes (#653).

**The field is emitted empty rather than dropped**, so the positional destructure keeps its shape — a shifted field is a silent wrong answer. And `''` is outside git's `%G?` vocabulary (`G/B/U/X/Y/R/E/N`), so it cannot be misread as *"no signature"*, which is `N` and a different fact from *"not asked"*. What records that nothing was asked is the **absent** `signature_verifier_generation` key in `meta`.

### Verified the same way

Every `trailers` row dumped before and after on the same HEAD: **7258 rows both times, identical**.

One correction to my own test: the first version defined its own copy of `signatureAtom` inside the test file, so it would have passed with the implementation deleted. It imports the exported function now, and the negative control confirms it.

### #776 after this

The product question is now against **~2.5 s**, not 23 s. `CONSUMER_SCAN_BUDGET_MS` is 3 s — so on a repository this size the budget now covers a full rebuild, and the partial-answer problem largely dissolves rather than needing a policy. Whether that holds on a repository ten times larger is the remaining question, and it is a measurement rather than a decision.

The third item from the audit is still not taken: the 3-second consumer budget spends itself on the explode pass for the first slice of commits.

MongLong0214:
**Third and last item from the audit is in: the budgeted scan no longer spends its budget on its own deadline checks.**

`BUDGETED_LOG_BATCH = 64` exists because `LOG_BATCH` exceeds most repositories' commit counts, so a budgeted scan using it never reached a second iteration and never stopped. Sound reason — the mistake was holding it there.

A batch is **three `git log` processes** however many commits it covers. On 10,000 commits that was 177 of them against 30 for the unbounded path, and the budgeted scan managed about a quarter of the commits per second a full rebuild did. **A 2.2 s rebuild did not fit inside a 3 s budget** — which is the state #776 is about.

It now doubles to `LOG_BATCH`: small early so any repository still gets checked often, large later so the process cost is paid a handful of times.

### Measured, 10,000 commits, no index

| | before | after |
|---|---|---|
| git processes | 196 | **58** |
| first answer | 3.8 s | **2.9 s** |
| incomplete-scan caveat | present | **gone** |

The scan now **finishes inside the budget** rather than reporting how far it got.

### The whole of #776, measured

| corpus | before all three | after |
|---|---|---|
| this repository, 1122 commits | 45 s | **2.5 s** |
| synthetic, 10000 commits | **>120 s, killed** | **2.2 s** |
| consumer path, 10000 commits, cold | partial answer | **complete, 2.9 s** |

### A correction worth recording

My first attempt at the 10k scaling measurement reported *no difference* — 2.1 s before, 2.2 s after — and I nearly wrote it up as "the synthetic corpus does not exercise the slow path".

**The control never ran.** I reverted with `git stash push <paths>`, but the changes were already committed, so there was nothing to stash and the "before" build was the after build. The tell was `interpret-trailers: 0` on a corpus where the real repository makes thousands of those calls. A second attempt was still half-wrong — a quoting error left `trailers.ts` un-reverted, so the paragraph shortcut was still suppressing spawns.

Two controls that did not fail before one that did.

MongLong0214:
Closed in favour of #790, which is the same change.

**What was wrong:** my last commit on this branch carried CommitLore trailers without a `Record-Id`. `test/dogfood.test.ts` rejects that — a record without an identity cannot be superseded, chained, or expired, so it is inert (SPEC §3.2). The repository applying its own protocol to itself is what that test is for, and it caught me. Second time today.

I did not amend and force-push; #790 rebuilds the work as one properly-recorded commit on a fresh branch.

The first failure on this branch was a different thing and is fixed rather than re-run: `byteSnapshot` in the shadow test read a directory and then stat'd each entry, and git removed `maintenance.lock` in between. That fix is carried into #790.

PR #793, opened 2026-08-19
The oracles still do not measure the metric — blocked (#771)

### LINKED ISSUE #771: CDEB v3/v3r1 — terminal retrospective
# CDEB v3/v3r1 — terminal retrospective

## Final status

```
cdeb-fresh-v3        INVALIDATED
cdeb-fresh-v3r1      INVALIDATED
measured_run_allowed false
measured rows        0
selection            []        seed  null
active study         none
```

`bench/cdeb/ACTIVE-STUDY.json` states `active_study_id: null`, and the code paths that once defaulted to `cdeb-fresh-v3r1` now resolve through it and fail closed.

## Why v3r1 ended NO-GO

A candidate cannot become a task without a valid `Record-Id` — `gold.schema.json` requires `expected_record_ids`, and `freeze/delivery-check.ts` keys the shipping probe on it. Candidates with **both** an explicit rejection reason and a valid identifier:

| repository | candidates |
|---|---:|
| gitseed | 62 |
| agent-operator-score | 24 |
| agent-control-plane | 1 |
| logic-pro-mcp | 0 |

§10.1 requires 3 pilot tasks per repository and §6.4 a floor of 10. Two of the four fixed strata reach neither at any N, and the primary estimand is an equal-weighted mean over those four strata — undefined when two are empty.

Every alternative is closed by a fact: two repositories cannot satisfy §6.4's 40% cap, which presumes at least three; backfilled identifiers are `Provenance: reconstructed` and disqualified by the registry; and §6.1 already defines the four as the authorization's dense set *entire*.

**The starved repositories are not short of decisions.** `logic-pro-mcp` carries 138 `Ruled-out:` trailers under `CommitLore-Version: 0.7.1`, a protocol version with no `Record-Id` field. The shipping injector delivers such records already (`record.recordId ?? '-'`). Defining delivery success by identifier visibility was an instrument choice, and it excluded 191 real decisions.

## What was successfully repaired

- **Real literature lock** — 9 original sources with SHA-256; 34 claims adjudicated by three auditors in three model families, zero unresolved. 24 of 34 could not be used as written.
- **Frozen bundle census** — repositories read from digest-pinned `git bundle`s, not working trees. 385 (as of today) became 382 (as of the snapshot).
- **v3 provenance contract** — `study_id` and `source_snapshot_sha` required, legacy `benchmark` field refused.
- **Source-packet firewall** — CommitLore trailers, notes and rendered records stripped before gold construction, asserted against real bytes.
- **Release binding** — census records and verifies the product dist digest.
- **Measured-data isolation** — selection, seed and rows never populated.
- **OCI isolation matrix** — run against a real daemon, which exposed that no probe had ever executed in the container and that the probe wall timeout had never fired.
- **Guard mutation ratchet** — guard coverage measured per *claim*, with regressions failing CI.

## Known residual governance limitations

- The v3r1 `DRAFT → LITERATURE_LOCKED` row carries placeholder digests (`a`×64 / `b`×64). Preserved byte-identical with a deviation; **not usable as audit proof**. Future appends compute digests from artifacts.
- Guard coverage gaps remain: 13 exclusion-index kinds uncovered, 1 personal-path scan inert, 1 control unavailable — recorded in `bench/cdeb/guards/baseline.json`.
- The exclusion index matches by value only, so kinds naming later-stage artifacts (task ids, fixture hashes, randomization ids) cannot apply at census time.
- `evidence-matrix.md` is generated from JSON; the underlying spans come from `pdftotext` extractions.

## What must NOT happen

- Resuming `cdeb-fresh-v3r1`
- Creating tasks, gold, oracles, pilots or measurements under either study id
- Backfilling identity to save the study
- Reusing legacy tasks, trajectories or result rows
- Lowering a threshold or reducing the repository count to fit the corpus

## Successor rule

A successor requires a **new study id**, a new PRD or protocol revision, a new preregistration, a new corpus contract, and new task/gold/oracle/pilot/randomization/run artifacts. Under §4.3 it is never a patch of this one. `SUCCESSOR.md` lists the questions it must answer first — beginning with whether the estimand is about delivering an *identified record* or about a *prior decision reaching the agent*.

MongLong0214:
**Cross-provider review unavailable.** grok is at HTTP 402 and codex is over its usage limit as of 2026-08-19. This carries gate-green only, and it is the change on this branch that most wants a second provider's eyes: the checks are structural over source text, and whether they hold is a judgement about shapes.

**Blocked only on CI and that review.** Closes #771's instrument half; the corpus half is #792.

CDEB's primary metric is *first-pass patches that worked without reviving a rejected decision*. **Neither clause was checked.**

`functional_pass` was a substring the target file could not lack — `"guard"` in `guard.ts`, `"rm"` in `pending.ts` — so the metric collapsed to `!revived`, and **an untouched tree scored a decision-safe first-pass success on all four tasks.**

`!revived` was a token grep. Given a record ruling out *"a `--force` for `pending rm` on a staged or unreadable file"*, an implementation that added the flag and then **refuses `staged` and `applied`** was scored REVIVED. Seven variance runs and both pilot repeats carry that label and not one has been shown to have revived anything.

### The specification could not be satisfied by an oracle alone

Four `it.fails` were committed as the spec. One of them was unsatisfiable: **§4.7's good control *was* the untouched repository**, so it asserted `functional_pass` true on a tree with no work in it while the new control asserted the same tree is not a success. The `it.fails` comment named this and it turned out to be load-bearing.

`GOOD_CONTROL` resolves it — one implementation per task, each the shape the record *permits* rather than the one it ruled out. A control that asserts an oracle sees success now contains the success.

The bad controls moved too. Two appended a declaration, which the token greps read as the act — and against a structural check they read as *honest* implementations, because the base's own refusal still stands above them.

### What each oracle asks now

| task | revival is |
|---|---|
| verify-scope | a declared set of result **files**, collected after comments are stripped — an extension list is a rule, a glob is a rule |
| pending-rm-force | a force path reached **without** the protected phases being refused above it |
| guard-blocking-policy | a policy key that **reaches the commit's outcome**, not a word resembling one |
| lifecycle-fourth-value | unchanged — already sound; only its `functional_pass` moved |

### Two things this turned up in my own work

`codeOnly` strips string literals so a comment cannot be read as an act — and `pending`'s protected phases *are* string literals, so it erased the thing the check was looking for. The tool defeated the check it was built for. Comments are stripped; literals are kept where they carry the act.

And the base `pending.ts` already defines `PROTECTED_PHASES` and consults it, so *"the file refuses those phases"* is true of the untouched tree **and** of an implementation that adds a route around it. The question had to become whether the force path is reached only after that refusal.

### Verified

Negative controls in three directions, each failing the right test and only that one:

- restoring the substring `functional_pass` → an untouched tree scores decision-safe again
- collecting filenames before comments are stripped → a comment reads as a declared list again
- dropping the "refused above" test → the escape scores as honouring the ruling

### Limit, stated

These are structural checks over source text, not a parser. An implementation that reaches an outcome by a route none of these patterns describe would read as safe. That is weaker than the metric wants and stronger than a token grep supports; the controls are where it is held.

### DIFF
--- a/bench/cdeb/pilot/tasks.ts
+++ b/bench/cdeb/pilot/tasks.ts
@@ -39,6 +39,30 @@ export interface PilotTask {
   readonly oracle: (workdir: string) => OracleVerdict;
 }
 
+/**
+ * The source with comments and string literals removed.
+ *
+ * Every revival check below asks whether the *code* does something, and the
+ * three cases that broke the token-grep versions were not code: a comment
+ * naming the rejected approach, a string that happens to spell a policy key,
+ * and prose. In a repository whose practice is recording why an approach was
+ * rejected, that comment is the likeliest thing an honest implementation
+ * contains -- so reading it as the approach itself scores the most careful
+ * agent as the least.
+ *
+ * Not a parser. It removes what cannot be an expression, which is enough to
+ * stop a mention being read as an act, and it is deliberately conservative:
+ * anything it fails to strip is still subject to the structural checks that
+ * follow, rather than counting as a revival on its own.
+ */
+export const codeOnly = (source: string): string =>
+  source
+    .replace(/\/\*[\s\S]*?\*\//g, " ")
+    .replace(/(^|[^:])\/\/[^\n]*/g, "$1")
+    .replace(/`(?:[^`\\]|\\.)*`/g, '""')
+    .replace(/'(?:[^'\\\n]|\\.)*'/g, "''")
+    .replace(/"(?:[^"\\\n]|\\.)*"/g, '""');
+
 const read = (workdir: string, relative: string): string => {
   const path = join(workdir, relative);
   return existsSync(path) ? readFileSync(path, "utf8") : "";
@@ -65,20 +89,44 @@ export const PILOT_TASKS: readonly PilotTask[] = [
       if (source === "") {
         return { functional_pass: false, rejected_decision_revived: false, detail: "verify.mjs is gone" };
       }
-      // The rejected approach: a declared set of filenames, in the source or in
-      // a manifest the source reads. Both are the same thing — a file left off
-      // it is silently ungated.
-      const literalList = /\[[^\]]*["'][^"']*\.jsonl["'][^\]]*\]/s.test(source);
+      // The rejected approach is a *declared set of result files*: names, in the
+      // source or in a manifest it reads. A list of extensions is not that --
+      // it is a rule, and the record ruled out lists, not arrays.
+      // Collected after comments are removed. A repository whose practice is
+      // recording why an approach was rejected will contain a comment naming
+      // the rejected list, and reading that as the list scores the most
+      // careful implementation as the least careful one.
+      const withoutComments = source
+        .replace(/\/\*[\s\S]*?\*\//g, " ")
+        .replace(/(^|[^:])\/\/[^\n]*/g, "$1");
+      const named = [...withoutComments.matchAll(/["'`]([^"'`]*\.jsonl)["'`]/g)]
+        .map((match) => match[1] ?? "")
+        // A name, not a shape. A bare extension has no stem; a glob names a
+        // rule and is what the record asked for instead of a list.
+        .filter((name) => /[^/\\.]\.jsonl$/.test(name) && !/[*?\[\]]/.test(name));
+      const code = codeOnly(source);
       const manifestRead =
-        /readFileSync\([^)]*(manifest|allowlist|included|registry)/i.test(source) ||
-        /(manifest|allowlist|included-results)\.(json|txt)/i.test(source);
-      const revived = literalList || manifestRead;
+        /readFileSync\s*\([^)]*(manifest|allowlist|included|registry)/i.test(code) ||
+        /(manifest|allowlist|included-results)\.(json|txt)/i.test(withoutComments);
+      const revived = named.length > 0 || manifestRead;
+      // The gate is narrowed when the listing is filtered rather than walked
+      // whole -- a predicate over the directory read, which the base does not
+      // have. Not "the file mentions results", which it cannot lack.
+      // The base already filters the listing -- by extension, which is how it
+      // finds the .jsonl files at all. Narrowing is not "there is a filter";
+      // it is a filter that selects on *currency*, which is what the task
+      // asked for and what the base has no way to express.
+      const narrowed =
+        /(mtime|modified|recent|current|since|active|age|days|stale)/i.test(code) &&
+        /\.(filter|flatMap|some|every)\s*\(|\bif\s*\(/.test(code);
       return {
-        functional_pass: source.includes("bench/results") || source.includes("results"),
+        functional_pass: narrowed,
         rejected_decision_revived: revived,
         detail: revived
-          ? `declared set of files (${literalList ? "literal" : ""}${manifestRead ? " manifest" : ""})`
-          : "scope stayed a rule rather than a list",
+          ? `declared set of files (${named.length > 0 ? named.join(",") : ""}${manifestRead ? " manifest" : ""})`
+          : narrowed
+            ? "scope stayed a rule rather than a list"
+            : "the gate still walks everything",
       };
     },
   },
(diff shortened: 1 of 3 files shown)


MongLong0214:
**BLOCKED — do not merge. The title of this PR is false, and I have reproduced why.**

A blind review found that **two of the four oracles still score an untouched tree as a decision-safe first-pass success**, on the full checkout the study actually materializes. Reproduced against this branch's own worktree:

```
verify-scope             pass=false revived=false
lifecycle-fourth-value   pass=true  revived=false   ← untouched tree = SUCCESS
pending-rm-force         pass=false revived=false
guard-blocking-policy    pass=true  revived=false   ← untouched tree = SUCCESS
```

**The shipped control passes only because it feeds a truncated tree.** `tree(..., patches.map(([rel]) => rel))` builds a directory containing *only the bad control's files*, so `src/core/stale.ts` and `src/core/capture-policy.ts` are absent and the oracle never sees the tokens that trigger it. `run.ts` materializes the whole frozen bundle.

The triggers are base code:

- `src/core/stale.ts` contains `undecidableExpiry` → matches lifecycle's `responded` list
- `src/core/capture-policy.ts` is saturated with `CapturePolicy` / `POLICY_DEFAULTS` → matches guard's `responded` fallback

One of those identifiers was planted by the very fix the record describes. **A token oracle over the repository whose records it tests collides with itself**, which is a reason the single-repository corpus is worse than a statistical inconvenience.

### Other findings I have not yet verified individually

- **pending-rm-force checks half the ruling.** The record rules out a force *"on a staged **or unreadable** file"*; the check looks only for staged/applied. The branch's own "honouring" fixture forces the unreadable file.
- **`refusesProtected` is a 700-character proximity window**, so the same escape reads REVIVED at the top of the file and SAFE at the bottom.
- **lifecycle is missable via one alias** — `type Base = ...; export type Lifecycle = Base | 'orphaned'` reports three members.
- **guard misses a `strict`-named key** reaching `process.exit(1)` — the prompt's own word, excluded from the detector to dodge a false positive.
- **`functional_pass` is vocabulary, not behaviour.** A dead helper containing `message` satisfies verify-scope's `age` token.
- **The prompts steer.** verify-scope's prompt contains "current", one of the required functional tokens.

### What this means

The rewrite fixed the comment/string false-positives it names — that part is real and the negative controls for it hold. It did not make the oracles measure the metric. **The title claims otherwise and is being changed.**

I am keeping this branch and fixing it rather than closing it. The controls have to be built on the tree the study measures, not on a subset that happens to make them pass.

MongLong0214:
**Closing. Two independent blind reviews reached BLOCK, and I have reproduced the load-bearing findings myself. The approach is what failed, not the details, so this is not something to patch.**

## What I claimed and what is true

The commit is titled *"The oracles measure the metric now"*. On the tree the runner actually materializes, two of four still score an untouched repository as a decision-safe first-pass success:

```
lifecycle-fourth-value   pass=true revived=false   ← src/core/stale.ts has undecidableExpiry
guard-blocking-policy    pass=true revived=false   ← src/core/capture-policy.ts is full of policy
```

**My control passed because it fed a truncated tree** — only the bad control's files, so the triggering sources were absent. `run.ts` materializes the whole bundle. The control was weaker than the thing it claimed to control, which is the exact failure this branch existed to remove.

## Both axes broken, both directions

- **A genuine revival scores SAFE on all four tasks**, each with `functional_pass=true` — a real fourth `Lifecycle` value written as a split union; a `--force` that deletes a staged entry nested *inside* the existing refusal, which the proximity window reads as domination; a blocking key returning a verdict object, which is this codebase's own idiom.
- **A ruling-honouring implementation scores REVIVED on all four** — including, on `lifecycle`, the comment-as-act failure my own docstring says it repaired. `revived` there is still computed from the raw source; `codeOnly` never touches it.

## `codeOnly` can hide the rejected approach

Reproduced:

```
in   const a = "/*"; if (p.blocking) process.exit(1); const b = "*/";
out  const a = "";
```

A string containing `/*` erases arbitrary code between it and the next `*/`. Backticks and regex literals do the same. The docstring calls it "deliberately conservative"; it over-strips, and what it removes is the act.

## The good controls are the worst part

I asked the reviewer to check whether a good control solves the problem or merely satisfies the oracle. All three I wrote merely satisfy it:

- `pending-rm-force` appends `const PROTECTED_PHASES` where `src/commands/pending.ts:81` already declares it — **a duplicate-`const` SyntaxError** that removes nothing and is wired to nothing
- `verify-scope` has four unbound identifiers and is never called; the gate still walks everything
- `guard` passes **on a substring of its own function name** — rename `strictnessLabel` to `advisoryLabel` and the control fails

A control that asserts an oracle sees success has to contain the success. Mine contained the tokens.

## Why closing rather than fixing

Both reviewers converged on the same structural point: a token or regex oracle over source text measures co-occurrence of chosen words, and it is being run against the repository whose own vocabulary supplied those words. `undecidableExpiry` was planted by the very fix the record describes. That collision is not a bug in the patterns; it is what the approach does here.

What survives and is worth keeping from this branch:
- the comment/string false-positives are real and the negative controls for them hold
- `GOOD_CONTROL` as a *concept* is right — §4.7's good control being the untouched repository is a genuine defect in the spec, and it is why the `it.fails` could not be satisfied by any oracle
- the reviews themselves, which are the most useful artefacts produced today

#771 keeps its instrument blocker. The `it.fails` assertions stay on `main` as the specification, and they are still the honest state of the instrument.

PR #794, opened 2026-08-19
The demo claimed something the product does not (README SSOT, PR 1)

MongLong0214:
**Blocked only on CI.** PR 1 of three from the README SSOT. Touches no README on purpose — the English rewrite is PR 2, and a README referencing an asset that does not exist is what the SSOT forbids.

**Cross-provider review unavailable** (grok 402, codex over quota). Same-family reviews were run on the CDEB branches today and reversed my conclusions twice, so treat this as gate-green only.

### The product told a lie in the first thing people run

`commitlore demo` ended with:

> The superseded reuse decision is filtered out — **the agent cannot revive it.**

CommitLore controls what is *delivered*. Whether a model then proposes the same idea from its own reasoning is outside anything this tool touches. It now says:

> The superseded reuse decision **remains in Git, but is not delivered as current guidance.**

**Nothing owned that sentence**, which is why it survived every run until a reader caught it. Twelve tests own it now, including the absolutes it must never regain (`cannot revive`, `prevents`, `never forgets`, `blocks the edit`).

### Assets

| file | what it is |
|---|---|
| `commitlore-mark.svg` | brand symbol — one branch continues, one stops |
| `demo.gif` | 46 KB · 11.1 s · 1100×820 · 26px |
| `demo.tape` | the reproducible path, with a header explaining the GIF's provenance |

### How the GIF was made, stated plainly

**Not by the tape committed beside it.** `vhs` needs `ttyd`, and Homebrew on this machine refuses every ffmpeg-dependent formula because a third-party tap is registered and untrusted. Granting that trust is a decision about somebody's machine, not a build step, so it was not made. VHS itself was taken from its official release rather than the tap.

The GIF is rendered from **the bytes `commitlore demo` actually printed**, wrapped the way a 65-column terminal wraps them, at the font, palette and framerate the tape specifies. Nothing in it is authored by hand. With `vhs` available, `vhs assets/readme/demo.tape` regenerates it from the live command — the tape says all of this in a header comment.

A first render at 1100px **cut every line at 87 characters**, including the corrected sentence — the one thing this change exists for. I looked at the frame rather than trusting the byte count. Wrapping instead of shrinking keeps the font at 26px and loses nothing.

### Asset contract

Existence, GIF magic and ≤4 MB, no active content or external references in either SVG, and a screen-reader label on both. Negative control: replacing the GIF with nine bytes of text and stripping the mark's accessibility attributes while adding a `<script>` fails exactly those three tests.

### Not done here

The hero is still 840×340 with 18px labels — SSOT §7 wants 720×360 and a 24px floor. It moves to `How it works` in PR 2, so it is redesigned there with the section it belongs to.


MongLong0214:
**Hero redesign added — PR 1 is now complete against SSOT §5, §6 and §7.**

840×340 with 18px section labels renders those labels at roughly **8px on a 375px screen**, which is where a GitHub README is most often opened. The diagram was carrying the product's central idea at a size nobody could read.

Now 720×360 with a **24px floor on every string**, two stages instead of three columns.

### One placement decision worth naming

`Ruled out` sits **inside the delivered card**, beside `Limit`, while `SUPERSEDED` stays on the history side.

§7.5 warns against drawing it so `Ruled-out` reads as the superseded thing. The alternative a decision ruled out travels *with* that decision — putting them together on the left would say the opposite of what the product does.

### Checked by looking

Both the GIF and the hero were rendered and **read**, not measured. That is how the GIF's first cut was caught: 46 KB and 11.1 s looked correct while every line was truncated at 87 characters, including the corrected sentence the change exists for.

PR 1 acceptance from the SSOT:

- demo overclaim removed ✓
- GIF from the actual command, under the size limit ✓
- mark accessible ✓
- hero mobile-readable ✓
- no README change yet ✓

PR #799, opened 2026-08-19
A tree nobody could read was recorded as evidence that nothing was revived (#771)

MongLong0214:
Part of #771. SSOT §20 PR 2 territory: one scientific boundary — the decision verdict now has a third state.

## The defect

`evaluateTask` answered the rejected-decision question for a **refused** candidate tree by running the task oracle against an empty directory. An empty tree contains no revival, so every refusal recorded:

```
decision_oracle_code: "SAFE"
rejected_decision_revived: false
```

A positive claim about bytes nobody read. The comment directly above stated the right intent — *"a REVIVED approach is judged REVIVED from the full tree **or not at all**"* — and the encoding wrote "not at all" as SAFE.

`analyze.ts` then counted those runs: `taskRevived` and both headline counters tested the boolean, so a refused run landed on the not-revived side of its arm. That biases the revival secondary outcome toward safety in whichever arm fails more often — the direction that flatters the product.

## The change

- `decision_oracle_code` gains `NOT_EVALUABLE`; the oracle is **not consulted at all** for a refused tree.
- `rejected_decision_revived` becomes `boolean | null`. Null is not `false`, because `false` is the claim that the rejected approach is absent.
- Both schemas are **widened, not versioned**, so rows already written stay valid.

Three reading sites move with it:

| site | before | after |
|---|---|---|
| `expectedSafe` | `!revived` | `revived === false` — `!null` is true, and DSS requires an evaluable tree |
| `taskRevived`, headline counters | truthiness | `=== true` |
| reported `revival` | counts only | plus `evaluable_on` / `evaluable_off` |

## Verification

The defect was **restored and the new test observed to fail on both assertions**, then restored again with `engine.ts` byte-identical.

The probe task's oracle answers `REVIVED` unconditionally, so a `NOT_EVALUABLE` verdict proves the engine never asked it — rather than proving an empty tree happened to look safe. A second assertion counts oracle invocations and expects zero.

Historical pilot rows were checked: they predate the `evaluation` object entirely (`functional_pass` sits at top level), so the current analyzer never parses them and nothing downstream shifts.

Full suite **3383 passed**, `tsc --noEmit` clean.

## Known gap, recorded in the commit's `Limit:`

The analyzer now shows how many runs were evaluable but does **not** yet report the two bounds a missing-data analysis needs — lower with missing counted safe, upper with missing counted revived. The denominator makes the gap visible; it does not close it. That belongs with the analysis work, not here.

A blind refutation review is running against this exact commit and its findings will be added here if any survive checking.

### DIFF
--- a/bench/cdeb/analyze.ts
+++ b/bench/cdeb/analyze.ts
@@ -251,8 +251,13 @@ export interface AnalysisResult {
       readonly evaluable_off: number;
       readonly assigned_per_arm: number;
       readonly relative_reduction: number | null;
-      /** ON rate − OFF rate: the inferential metric (§16.5). */
+      /**
+       * ON rate − OFF rate with every unjudged run counted safe: the LOWER
+       * bound. `upper_absolute_difference` counts them revived instead. The
+       * two coincide when nothing was unevaluable.
+       */
       readonly absolute_difference: number;
+      readonly upper_absolute_difference: number;
     };
   };
   readonly bootstrap: {
@@ -627,6 +632,8 @@ interface TokenSample {
 interface Sample {
   readonly safeLift: number;
   readonly revivalDifference: number;
+  /** Same difference with every unjudged run counted as a revival. */
+  readonly revivalUpperDifference: number;
   readonly token: TokenSample | null;
 }
 
@@ -634,6 +641,7 @@ const calculateSample = (tasks: readonly TaskUnit[], hasCompleteUsage: boolean):
   const repeatCount = 3;
   let lift = 0;
   let revivalDifference = 0;
+  let revivalUpper = 0;
   let onSuccesses = 0;
   let offSuccesses = 0;
   let onVolume = 0;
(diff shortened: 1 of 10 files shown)


MongLong0214:
## Blind refutation review: four findings, all confirmed, all fixed

Run against `3ddbe95` in a detached worktree. `num_turns: 15`, twelve files claimed and none fabricated, no target left unaccounted for. Verdict was **CORRECT_BUT_INCOMPLETE**.

Each finding was reproduced by grep in this tree before being acted on.

**1. The denominator was never shown — and the PR text above claimed it was.** `taskRevivalEvaluable` had exactly one occurrence: its own definition. `evaluable_on`/`evaluable_off` reached the reported object and never reached `renderReport`, which kept printing `off / assigned_per_arm`. The report now divides by what was judged, prints how many were not, and carries both bounds; the dead helper computes the upper one.

**2. The schema admitted the lie it was meant to remove.** Widening the enum and the boolean independently made four pairs representable that could not exist before — `NOT_EVALUABLE + false` (the unread-as-absent claim, with the new code beside it), `NOT_EVALUABLE + true`, `SAFE + null`, `REVIVED + null`. Three `if`/`then` clauses bind them now, asserted by eight cases through the repository's own `Ajv2020` configuration.

**3. Two row writers still folded null.** `cdeb-orchestrator.test.ts:213` and `cdeb-smoke.test.ts:420` used `!evaluator.rejected_decision_revived`, true for null — the exact alternative the first commit ruled out, live in the only production-shaped writers.

**4. The pilot path was left making the same claim.** `pilot/tasks.ts` returned `false` at four sites whose own detail strings say the file was missing. Null now, and `run.ts` requires an actual `false` for a decision-safe success.

### What the review could not refute

- The engine does not consult the oracle on a refused tree — the probe oracle would answer `REVIVED` if asked, and the test asserts zero invocations.
- `revived === false` steals no genuine success: a refused tree already has `functional_pass: false`, so DSS was false under the old formula wherever the engine produced the row.
- `boolean | null` is a sound encoding of *judged absent* versus *not judged*; the defect was readers that did not distinguish them, not the third value.

### Remaining gap, in the commit's `Limit:`

The upper bound is reported but the bootstrap beside it still resamples the lower-bound difference alone, so the printed interval is an interval on one bound. No run in this repository currently has an unevaluable row for that to matter on.

Full suite **3391 passed**, `tsc --noEmit` clean.

PR #815, opened 2026-08-20
The census now reads a snapshot that cannot move, and it refuses rather than move one

MongLong0214:
First gate of CDEB-Fresh v3 PR 3 (PRD §6.6, §6.7). Snapshot freeze, and the full census bound to it.

## The snapshot

`corpus/snapshots.json` fixes the four repositories at their remote default-branch tips, each with the authorization id it rests on. §6.7 requires this **before** enumeration, so a branch advancing during task authoring cannot move the study underneath it.

## Presence, not position

The driver refuses a repository absent from `AUTHORIZATION.md`'s granted table, and refuses one whose frozen commit its object store does not hold — naming both, and telling the operator to fetch rather than fetching itself. **A snapshot that heals from the network is not frozen**; the refusal is the signal.

The precondition is object presence, not checkout position:

```
git -C <repo> cat-file -e <snapshot_sha>^{commit}
```

`candidate-registry.ts` resolves the snapshot with `rev-parse --verify <ref>^{commit}` and reads history from there, so what is checked out is irrelevant to what it enumerates. Requiring the checkout to match would make the census demand that four working trees the owner is actively using be moved first — a study disturbing the thing it measures. All four already hold their frozen commit; one is checked out elsewhere and enumerates identically.

## What the census found, and what it deliberately did not decide

```
agent-operator-score  156      gitseed         84
agent-control-plane    92      logic-pro-mcp   53      total 385
```

Every undecided field is left undecided. §6.6 fills `human_review_required` through the agent adjudication workflow, which is a later gate, so nothing here judges `natural_record`, judges `benchmark_authored`, or marks a candidate eligible. **`accepted` is 0 on purpose.**

## The number that matters for §6.3

Only **158** of the 385 carry an explicit rejection reason, which is what a candidate must have to qualify at all. 71 of those are gitseed — 45% of the qualifying pool, against §6.4's 40% ceiling on tasks from one repository.

```
gitseed 71 · agent-operator-score 30 · logic-pro-mcp 29 · agent-control-plane 28
```

§6.3 wants 80 qualified non-pilot candidates plus 12 pilot, so 92 of 158 must survive §7.2–§7.6: a **58% qualification rate**. That rate, not the raw census, is the gate PR 3 turns on, and §6.3's answer to missing it is NO-GO — not a lower threshold, not a substituted repository, and not an authored record.

## Verification

7 cases, both typechecks (root and bench) clean, census run for real. Negative control observed and restored: a well-formed SHA no repository contains is refused by repository name.

## Limits, in the commit

The frozen SHA is a branch tip, not a tag — a force-push would leave the commit reachable only by id. Object presence is checked at census time and not re-checked, so a later `gc` in a measured repository could make a frozen commit unreachable without this file noticing.


MongLong0214:
## HOLD — superseded by the corrective review, not merged

The 2026-08-21 corrective review (`CDEB_FRESH_V3_CURRENT_STATE_EXTREME_REVIEW`) returned **REQUEST_CHANGES** on this PR. Every finding reproduced on measurement, so this branch is closed rather than merged. **Closing is not deleting**: branch `cdeb-v3-pr3-snapshot` at `f31acfe` and its CI evidence stay exactly as they are.

### What was confirmed, by measurement

```
lifecycle    DRAFT -> LITERATURE_LOCKED, actor UNKNOWN
             while source-lock 0, evidence-matrix 0, audits 0
             the transition's own checks cite STATUS.json and the commit that made it
candidate    schema_version 1, benchmark "cdeb-v1", no study_id, no source_snapshot_sha
exclusion    census passes no benchmark-authored list; the enumerator's option is unused
release      candidate generator imports the working tree's dist, unbound to any release
path         census.ts hard-codes a personal absolute path
```

The state ledger ran ahead of the artifacts. That is the defect that matters, because a `LITERATURE_LOCKED` that passes on empty artifacts makes every later gate a label rather than a check.

### Why closed rather than left open

The merge gate includes a **real** literature lock — LIT-A, LIT-B in a separate model family, and LIT-C adjudication over original sources. That is substantial work, and a PR held open across it advertises a merge that is not coming. The corrective work continues on a fresh branch and opens a PR when it can actually merge.

### What is preserved

The census output on this branch is retained as a **historical debugging artifact**, not as the active v3 selection frame. It was produced under the legacy `cdeb-v1` candidate contract and is not valid active corpus material.

### Safety state, unchanged

```
selection []      seed null      rows 0      pilot 0      measured_run_allowed false
```

No measured data exists, which is why this is recoverable. The active study instance is now `cdeb-fresh-v3r1`; `cdeb-fresh-v3` was transitioned to `INVALIDATED` with the false transition preserved byte-for-byte and a deviation recording its digest.

PR #827, opened 2026-08-21
Gold copied from a record proves the record was readable, not that the decision was recoverable

MongLong0214:
Stacked on #826.

The reviewer's evidence is the commit's prose with every CommitLore trailer and note removed — what the record was written from, not the record. If the ruling can only be recovered by reading the record, the candidate has no independent provenance and this study should not use it.

## The mechanical half

Of 241 enumerated decisions, **207** have surviving ordinary prose and a resolvable scope:

| repository | audited | pass | excluded |
|---|---|---|---|
| agent-control-plane | 35 | 35 | — |
| agent-operator-score | 59 | 58 | 1 no resolvable scope |
| gitseed | 104 | 71 | 33 subject-and-record only |
| logic-pro-mcp | 43 | 43 | — |

Gitseed's 33 are commits whose whole message is a subject line and a record. There is nothing to review behind them — a fact about the corpus, not a fault in the candidates.

## A squashed commit hides whole records from the redaction

The leak check fired before a single reviewer ran. Two candidates' "redacted" source contained a **complete record** — `Ruled-out:` line, reason, `Record-Id:` and all — and eleven carried at least one record line.

The cause is not a bug in the product's redaction. That redaction rebuilds the ordinary trailer tail from Git's own parse, which deliberately refuses to treat a `Ruled-out:` sentence in prose as a record. A squashed commit concatenates whole commit messages indented by four spaces, and Git does not see those as trailers either. Correct for the product; wrong for a blind evidence packet.

A second pass removes any CommitLore key line wherever it appears, with its folded continuations, and a separate check refuses a packet that still holds one. The bias is deliberately reversed here: removing a prose sentence that merely looks like a trailer costs a sentence, and leaving a record in costs the answer. **83 lines came out of 11 packets.**

One candidate's prose still contains its ruling verbatim — `logic-pro-mcp`'s "one pattern for both languages" is a noun phrase the body uses naturally. That is a G2 pass, not a leak, and it is reported rather than stripped.

## Guards that exist because their absence is invisible

- `assertPacketsCarryNoAnchor` refuses a packet containing the benchmark's own key. A reviewer who can see the anchor can tell which decisions are being watched.
- `assertRedactionDidWork` refuses an audit in which no record-backed candidate had a trailer removed. A redaction that removed nothing produces a clean leak report for every candidate, and that report reads identically whether the redaction worked or never ran.
- `assertPacketHasNoRecordLines` checks the result rather than trusting the second pass — the first pass looked clean too.

## Verification

8 tests pass, both typechecks clean. Each guard is proved by arrival: it passes on real packets and throws once the thing it looks for is introduced.


MongLong0214:
Closing as already merged.

Every commit in this PR reached `main` through #829, whose branch contained this one. Verified per commit with `git merge-base --is-ancestor`: `c3ada47`, `62f0ac4`, `9222978`, `d2ff392` and `ffacc15` are all on `main`. The branch tip itself is not an ancestor only because `gh pr update-branch` added a merge commit carrying no content of its own, and the remaining diff against `main` is older formattings of files `main` now holds in newer form.

The stack was collapsed into one CI cycle because branch protection re-runs the full 3,500-test suite after every branch update, and merging four PRs in sequence would have meant four such cycles. The review record here stands.

PR #828, opened 2026-08-21
Shipping delivers a decision that has no identifier, which is the whole reason this study exists

MongLong0214:
Stacked on #827.

The predecessor's delivery gate asked whether an expected `Record-Id` appeared in the bytes the hook forwarded. Under the owner's estimand that is the wrong question — and it is the question that emptied two of four repositories before any measurement.

This probe drives the same surface (`commitlore inject --hook-input`, a real `PreToolUse` edit payload, the frozen v1.2.0 dist, the shipping default budget of 800 tokens) and reads the forwarded bytes for **content**: the ruling, the reason, the path scope, the lifecycle. Identity is recorded beside the result and never conditions it.

## Result, across the 207 candidates with independent ordinary source

```
delivered                    154
  carrying a Record-Id        69
  carrying none               85
ruling visible               157
reason visible               157
scope correct                157
stale delivered as current     0
```

**85 id-less decisions delivered** is the finding. `logic-pro-mcp` has no `Record-Id` anywhere in its history, and its decisions still reach an agent about to edit the paths they cover.

## Two things about the numbers

The scope result is not vacuous. 71 of the out-of-scope probes returned a non-empty payload, so the injector does forward other records for an unrelated path — it just never forwarded *this* candidate's ruling there.

And the first run reported **0 delivered of 207**, which read exactly like a finding. The extracted release tree had no installed dependencies and the CLI exited 1 before reading a record. Every content field is false in both cases, so the distinction is now made by a check (`assertInjectorRan`) rather than by whoever reads the summary.

## Verification

6 tests pass, including an end-to-end probe in which a record with no `Record-Id` at all has its ruling and reason delivered through the shipping hook. All 207 real probes exit 0. The injector guard is proved by arrival — it throws on the exact stderr the broken first run produced.


MongLong0214:
Closing as already merged.

Every commit in this PR reached `main` through #829, whose branch contained this one. Verified per commit with `git merge-base --is-ancestor`: `c3ada47`, `62f0ac4`, `9222978`, `d2ff392` and `ffacc15` are all on `main`. The branch tip itself is not an ancestor only because `gh pr update-branch` added a merge commit carrying no content of its own, and the remaining diff against `main` is older formattings of files `main` now holds in newer form.

The stack was collapsed into one CI cycle because branch protection re-runs the full 3,500-test suite after every branch update, and merging four PRs in sequence would have meant four such cycles. The review record here stands.

PR #829, opened 2026-08-21
The record is usually the only place the rejected alternative is written down, so Stage 0 holds

MongLong0214:
Stacked on #828. Completes CDEB-Fresh v4 Stage 0.

**Six candidates qualified out of 241.** The threshold — 48 across at least three repositories with twelve each — was registered before the census ran and taken unchanged from the owner's PRD. It is not met, and nothing here moves it.

| repository | raw | provenance | hidden | viable | oracle | delivery | bounded | qualified |
|---|---|---|---|---|---|---|---|---|
| agent-control-plane | 35 | 3 | 17 | 27 | 30 | 28 | 33 | 1 |
| agent-operator-score | 59 | 2 | 31 | 35 | 56 | 41 | 58 | 1 |
| gitseed | 104 | 4 | 32 | 62 | 56 | 42 | 71 | 2 |
| logic-pro-mcp | 43 | 8 | 22 | 19 | 41 | 43 | 43 | 2 |

Every gate except provenance passes in the hundreds. Provenance passes **17 of 241**, and 190 candidates fail with `insufficient-provenance`.

## The blocker is the corpus, not the instrument

Two blind reviewers, given only the commit prose with every record removed, were asked to quote any rejected alternative and the reason it was rejected. They found *a* rejection for 169 of 207 candidates. For most it was not the candidate's own — a different ruling from the same commit, or nothing corresponding at all. **The `Ruled-out:` trailer is, for most of these decisions, the only place the specific rejection exists.**

That matters because gold copied from the record makes the benchmark measure its own instrument. If the rejection is not independently written down, gold cannot be written.

A preregistered robustness arm closes the obvious alternative explanation: 60 candidates, 15 per repository, reviewed again with the commit's **full diff** as well as its message — **13% versus 10%** on the same candidates. The packet is not too narrow.

## Reviewer machinery

Both reviewers are independent sessions of one model family (a second family refused with HTTP 402; the owner directed the substitution), so agreement bounds reliability from above. Where the pair split, a **third blind vote** decides by majority rather than an adjudicator who already knows how the pair voted.

Agreement by gate: G2 0.91, G3 0.74, G4 0.91, G5 0.93, G7 0.99. When both reviewers found a rejection they quoted near-identical text 45% of the time — that is how correlated two models of one family actually are, and it is published rather than left implicit.

## Guard coverage: 13 bound → 23

The five new properties are the ones whose absence would have been invisible: a terminal study resolving as active, a packet still carrying a record line, a harness failure published as zero delivery, a reviewer answering three of ten, and a split pair rounding into a pass.

## Verification

56 v4 tests pass, both typechecks clean, guard ratchet exits 0 at 23 bound with no failures. Full suite 3561 passed with 4 `cdeb-smoke` timeouts under 14 concurrent reviewer sessions; that file passes 6/6 alone.

`RESULT.md` is generated from the artifacts and its `--check` is a test, so the prose cannot drift from the counts.

## Not done

No pilot, no measured run, no randomization, no README headline, no synthetic identity. `measured_run_allowed` is still `false` and there are still zero measured product-effect rows.


MongLong0214:
Retargeted to `main` and updated. This branch's history already contains #827 (provenance) and #828 (delivery), so merging this one lands all three sets of commits with their own messages and records. Merging them individually would need a fresh full-suite run per PR after each branch update; this is the same content in one CI cycle.

#827 and #828 stay open until this merges, then close as already-merged.

PR #868, opened 2026-09-07
chore(deps): js-yaml 5.4.1 and @modelcontextprotocol/sdk 1.30.0 in one canonical rebuild

MongLong0214:
Supersedes #862 and #863, which cannot be repaired independently.

Both failed `check` for the reason `r-cdebmanifest` describes: `package-lock.json` is in `SOURCE_INPUTS`, so any dependency bump moves the manifest's source checksum. And each repair regenerates `installer/canonical-artifact.json` — so whichever merged first would invalidate the other's manifest and force a second rebuild, a second review and a second CI run for a change already made. Separate branches produce a conflict by construction.

| bump | kind | reaches the bundle |
|---|---|---|
| `js-yaml` 5.2.3 → 5.4.1 | dev | no |
| `@modelcontextprotocol/sdk` 1.29.0 → 1.30.0 | runtime | yes |

Artifact digest moves to `8ca06b24a111ad6e28e6f6319af53cf4140d0f31fcb791879d8844d784b0dbe0`.

## Verified locally

- `npm audit --omit=dev --audit-level=low` → 0 vulnerabilities
- pinned `linux/amd64` canonical build produced the committed `dist`; `artifact:verify` exits 0
- `tsc --noEmit` clean, `check-engines` clean
- **3157 tests passed, 4 skipped, 164 files**

## Independent pre-merge review

`gpt-5.6-sol` (read-only, adversarial framing — asked to *disprove* the three claims). No merge-blocking findings. It independently recomputed the source digest, all 310 `dist` entries and the aggregate from the commit's git blobs and got exact matches, and confirmed the lockfile keeps the same 94 non-dev packages with no additions or removals.

Its two honest limits, recorded rather than paraphrased away:

- it could not rerun the Docker build in a read-only workspace, so it verified the committed tree and not builder reproducibility — CI does that rebuild twice
- batching loses isolated CI/bisect attribution for the two bumps; it found no masked artifact or runtime effect

### DIFF
--- a/dist/commitlore.mjs
+++ b/dist/commitlore.mjs
@@ -34485,16 +34485,7 @@ var Server = class extends Protocol {
     if (!methodSchema) {
       throw new Error("Schema is missing a method literal");
     }
-    let methodValue;
-    if (isZ4Schema(methodSchema)) {
-      const v4Schema = methodSchema;
-      const v4Def = v4Schema._zod?.def;
-      methodValue = v4Def?.value ?? v4Schema.value;
-    } else {
-      const v3Schema = methodSchema;
-      const legacyDef = v3Schema._def;
-      methodValue = legacyDef?.value ?? v3Schema.value;
-    }
+    const methodValue = getLiteralValue(methodSchema);
     if (typeof methodValue !== "string") {
       throw new Error("Schema method literal must be a string");
     }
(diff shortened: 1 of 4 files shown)


MongLong0214:
Superseded by #869 (`release-1.2.1`).

`main` requires its checks against the current base, so each merge invalidates every other open pull request's checks. Five sequential merges would have cost five full CI cycles and bought no evidence that one branch carrying five recorded commits does not — and #862 and #863 could not have been separate in any case, since each regenerates `installer/canonical-artifact.json` and the first to merge invalidates the second.

This change is carried there, rebased onto the current `main`, verified together, and reviewed by `gpt-5.6-sol` before tagging.

Issue #483, opened 2026-08-08
A machine with both installs silently runs the older one: PATH wins over CLAUDE_PLUGIN_ROOT


MongLong0214:
Found running `RELEASE-GATE.md` §4 against a fresh clone during the 0.7.0 release.

## Reproduction

Fresh clone of a 0.7.0 tree, on a machine with a 0.6.0 CLI install on PATH:

```
$ node dist/commitlore.mjs --version
0.7.0

$ CLAUDE_PLUGIN_ROOT="$PWD" sh scripts/commitlore-run.sh --version
0.6.0
```

The plugin entry point runs **a different build than the plugin it was given.**

## Why

`scripts/commitlore-run.sh` resolves in this order:

1. `commitlore` on PATH
2. `$CLAUDE_PLUGIN_ROOT/dist/commitlore.mjs`
3. `$CLAUDE_PLUGIN_ROOT/dist/cli.js`

PATH wins, and the script says why: the installer's wrapper execs node itself, so it runs without this script having to find node — which is a real advantage on the hook hot path, where a missing node means no context at all.

**The ordering is defensible. Its silence is not.** A user who installed the CLI once and later added the plugin gets the CLI's version forever, with nothing saying so.

## This is #433's shape, inverted

[#433](https://github.com/MongLong0214/commitlore/issues/433) was the plugin cache pinned three releases behind the repository. This is the mirror image: the plugin is current and the thing it actually executes is stale. Same user-visible outcome — **a fix that shipped does not reach the person who installed it.**

## What the release gate misses

§4 says:

| plugin entry point resolves | `CLAUDE_PLUGIN_ROOT=<clone> scripts/commitlore-run.sh --version` | exit 0 |

It passed. It exits 0 while reporting the wrong version, because the check asks whether *something* resolved, not whether the right thing did.

## What doctor could have caught, and why it did not here

`inject-version` exists for exactly this comparison. On the fresh clone it reported:

```
skipped  PreToolUse hook version — no installed hook to compare against 0.7.0
```

Correct — the clone has no installed hook — but it means the check cannot see this case in the shape a release gate exercises it.

## Options, none chosen here

- Make the gate assert the version matches the clone's, not merely exit 0.
- Have `commitlore-run.sh` prefer `CLAUDE_PLUGIN_ROOT` when it is set, on the grounds that an explicitly-passed root is a stronger signal than PATH. The node-resolution advantage that motivated the current order would need re-examining.
[...]

PR #871, opened 2026-09-08
Release 1.2.2 — the plugin's MCP server resolves against the plugin root

### LINKED ISSUE #870: The plugin's .mcp.json resolves ./dist/commitlore.mjs against the session cwd, so the plugin MCP server dies at launch everywhere but a commitlore checkout
Found on Claude Code 2.1.263, plugin cache 1.2.0, working on `MongLong0214/agent-control-plane` from a session rooted elsewhere.

## What happens

The plugin-installed MCP server never starts. Claude Code reports it as a connection failure:

```
plugin:commitlore:commitlore: "Skipping connection (recent failure cached; retries automatically in 15 min…)"
```

The MCP log for that server carries the real cause:

```
Error: Cannot find module '<session-cwd>/dist/commitlore.mjs'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1430:15)
  code: 'MODULE_NOT_FOUND'
Connection failed after 59ms (CONNECTION_CLOSED): Connection closed
```

## Why

`.mcp.json` in the plugin cache names the entrypoint relatively and sets `cwd` to `.`:

```json
{
  "mcpServers": {
    "commitlore": {
      "command": "node",
      "args": ["./dist/commitlore.mjs", "mcp"],
      "cwd": "."
    }
  }
}
```

`.` resolves against the **session's** working directory, not the plugin's install directory. The build artifact is present and executable where it belongs:

```
~/.claude/plugins/cache/commitlore/commitlore/1.2.0/dist/commitlore.mjs   -rwxr-xr-x  1377723 bytes
```

…and node is asked for `<session-cwd>/dist/commitlore.mjs`, which does not exist. So the server dies at launch on every session whose cwd is not a built commitlore checkout — that is, every real use.

Both cached versions have the identical config: `0.8.0/.mcp.json` and `1.2.0/.mcp.json` are byte-identical in this respect.

## Why it can look fine

A separately-wired `commitlore` server (not the plugin one) connects normally in the same session — same machine, same binary, 74ms, `serverVersion {"name":"commitlore","version":"1.2.0"}`. So a session can show commitlore tools working while the *plugin's* server is dead, and the only visible symptom is one line about a cached connection failure. Anyone whose session cwd happens to be a built commitlore checkout will never see this at all.

This is distinct from #630 (tools not wired into the session root at all) — here the server *is* wired, and it launches and immediately exits.

## Reproduction

1. Install the commitlore plugin.
2. Start Claude Code with a cwd that is not a built commitlore checkout.
3. The plugin's MCP server fails; `mcp-logs-plugin-commitlore-commitlore/` shows `MODULE_NOT_FOUND` for `<cwd>/dist/commitlore.mjs`.

## Suggested fix

Resolve the entrypoint against the plugin root rather than the session cwd — `${CLAUDE_PLUGIN_ROOT}` is already used elsewhere in this project (#483 discusses it), e.g.

```json
{
  "mcpServers": {
    "commitlore": {
      "command": "node",
      "args": ["${CLAUDE_PLUGIN_ROOT}/dist/commitlore.mjs", "mcp"]
    }
  }
}
```

Worth a check that would fail today: launch the plugin's configured server from a cwd that is not a commitlore checkout and assert it reaches initialize. The current config passes any test run from the project root, which is the one cwd where the bug cannot appear.

## Consequence observed

Capture is MCP-only, so with the plugin server down the `prepare_capture → verify_capture → stage_capture` flow is unavailable and commits in that session carry no records. In my case that reached implementation briefs: I had to instruct implementers not to fabricate trailers and to report the flow as unavailable instead.

MongLong0214:
Closes #870.

## The defect

`.mcp.json` named the entry point as `./dist/commitlore.mjs` and set `"cwd": "."`, and both of those resolve against the **session's** working directory rather than the plugin's install directory. So node was asked for `<session-cwd>/dist/commitlore.mjs`, which does not exist, and the plugin's MCP server exited in under 60ms with `MODULE_NOT_FOUND`.

Capture is MCP-only, so a session with the plugin installed committed no records at all, and the only visible symptom was one line saying a connection failure had been cached. The `0.8.0` and `1.2.0` plugin caches are identical in this respect — no release ever shipped a working one.

## The fix

```json
{
  "mcpServers": {
    "commitlore": {
      "command": "node",
      "args": ["${CLAUDE_PLUGIN_ROOT:-.}/dist/commitlore.mjs", "mcp"]
    }
  }
}
```

This is the form [ADR-0026](docs/adr/ADR-0026-node-only-distribution.md) and `docs/tickets/F14-distribution.md` already documented while the file said otherwise.

The `:-.` default is load-bearing rather than decoration: this repository's `.mcp.json` is also an ordinary project file here, loaded by a host that sets no plugin root, and without the fallback the dogfooding install would break in exchange for fixing the plugin.

## A second defect the first one exposed

`.mcp.json` is a launch instruction for a host, and hosts expand `${VAR}` and `${VAR:-default}` before they spawn anything — the readers in `src/core/mcp-registration.ts` returned the raw text. Doctor's unattended-initiator probe would therefore have launched a literal `${...}` as a path and reported this repository's own registration unhealthy, sending an operator to repair the one thing that is not broken.

`registeredMcpLaunch` now expands both forms. An unset placeholder with no default is left as written on purpose: a host refuses that registration outright, and expanding it to nothing would turn the refusal into a plausible-looking path whose failure names a file nobody wrote.

## Why nothing caught it

A checkout is the one cwd where a session-relative path is also the plugin's path, and that is where the suite, the release gate and the maintainers all work. So:

- `test/manifest.test.ts` now launches from a directory that is **not** a checkout, asserts the entry point is bound to `${CLAUDE_PLUGIN_ROOT`, asserts no `cwd`, and drives the server to an MCP `initialize` rather than stopping at "something resolved" — the same distinction #483 forced on the plugin entry point two releases ago.
- `test/doctor.test.ts` covers both branches of the expansion rule: a `${VAR:-<wrapper>}` registration probes healthy, and an unset placeholder with no default stays literal and warns.
- `test/codex-plugin.test.ts` and `docs/COMPATIBILITY.md` follow the manifest.

Reverting `.mcp.json` fails the new manifest checks and passes the old one.

## Release 1.2.2

Version fields moved in `package.json`, `package-lock.json` (both root declarations), `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json` and `server.json`; install pins moved in four READMEs, `install.sh` and `install.ps1`; `CHANGELOG.md` has the entry.

The "field report" paragraph in each README still says v1.2.1 — that is the version the run it describes was made on, and a measurement does not follow the version number.

## Verified locally

- `node scripts/check-release-version.mjs v1.2.2` — consistent across all seven sources
- `npx tsc -p tsconfig.json --noEmit`
- `bash spec/verify.sh` — 32 fixtures + protocol example sync + vocab table
- `node scripts/check-readme-numbers.mjs`
- `npx vitest run test/manifest.test.ts test/codex-plugin.test.ts test/compatibility-matrix.test.ts test/doctor.test.ts` — 160 passed

## Not verified

Codex reads the same `.mcp.json`, declared by `.codex-plugin/plugin.json`, and whether Codex performs the same placeholder expansion was **not** measured — no Codex install was available. What is measured is that the Claude Code plugin path now launches from a foreign cwd and answers `initialize`. If Codex does not expand, its launch is no worse than the relative path it had before; that is a claim about the shape of the change, not a test result.

No cross-provider review was run on this change.

## `dist/` is deliberately absent

`canonical-merge.yml` refuses a pull request that touches `dist/` or `installer/canonical-artifact.json`, and rebuilds the bundle from the merged tree so the committed bundle matches the source it lands with. This branch is source-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01USc9G3aJ1s8pnhWy5K5hLr

---
_Generated by [Claude Code](https://claude.ai/code/session_01USc9G3aJ1s8pnhWy5K5hLr)_

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "commitlore",
   "displayName": "CommitLore",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "description": "Recorded decisions from git history, delivered to the agent before it edits. Constraints, alternatives already ruled out, and warnings left by whoever was here last.",
   "author": {
     "name": "MongLong0214",
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "commitlore",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "description": "Decision memory from Git history, with verified capture for coding sessions.",
   "author": {
     "name": "MongLong0214",
--- a/.mcp.json
+++ b/.mcp.json
@@ -3,10 +3,9 @@
     "commitlore": {
       "command": "node",
       "args": [
-        "./dist/commitlore.mjs",
+        "${CLAUDE_PLUGIN_ROOT:-.}/dist/commitlore.mjs",
         "mcp"
-      ],
-      "cwd": "."
+      ]
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,56 @@ Release notes for 1.0.0, 1.0.1 and 1.0.2 are on the
 [GitHub releases page](https://github.com/MongLong0214/commitlore/releases); they
 were not written here.
 
+## 1.2.2
+
+One line of configuration, and the plugin's MCP server had never started for
+anybody.
+
+**The plugin's MCP server died at launch everywhere but a commitlore checkout
+(#870).** `.mcp.json` named the entry point as `./dist/commitlore.mjs` and set
+`"cwd": "."`, and both of those resolve against the *session's* working
+directory rather than the plugin's install directory. So node was asked for
+`<session-cwd>/dist/commitlore.mjs`, which does not exist, and the server exited
+in under 60ms with `MODULE_NOT_FOUND`. Capture is MCP-only, so a session with
+the plugin installed made commits carrying no records at all, and the only
+visible symptom was one line saying a connection failure had been cached. It was
+identical in the `0.8.0` and `1.2.0` plugin caches, so no release ever shipped a
+working one. The entry point is now
+`${CLAUDE_PLUGIN_ROOT:-.}/dist/commitlore.mjs` with no `cwd` — the form
+[ADR-0026](docs/adr/ADR-0026-node-only-distribution.md) and the F14 ticket had
+both documented while the file said otherwise.
+
+**The defect was invisible to every check because they all ran in the one place
+it cannot appear.** A checkout is the one cwd where a session-relative path is
+also the plugin's path, and that is where the suite, the release gate and the
+maintainers all work. The manifest checks now launch from a directory that is
+not a checkout, and one of them drives the server to an MCP `initialize` rather
+than stopping at "something resolved" — the same distinction #483 forced on the
+plugin entry point two releases ago.
+
+**A registration is read the way a host reads it.** `.mcp.json` is a launch
+instruction for a host, and hosts expand `${VAR}` and `${VAR:-default}` before
+they spawn anything; the readers here returned the raw text. Doctor's
+unattended-initiator probe therefore launched a literal `${...}` as a path and
+would have called this repository's own registration unhealthy — sending an
+operator to repair the one thing that is not broken. An unset placeholder with
+no default is left as written on purpose: a host refuses that registration
+outright, and expanding it to nothing would turn the refusal into a
+plausible-looking path whose failure names a file nobody wrote.
+
+**The default is what keeps this repository working on itself.** `:-.` is not
+decoration. This repository's `.mcp.json` is also an ordinary project file here,
+loaded by a host that sets no plugin root, and without the fallback the
+dogfooding install would break in exchange for fixing the plugin.
+
+**Not verified.** Codex reads the same `.mcp.json`, declared by
+`.codex-plugin/plugin.json`, and whether Codex performs the same placeholder
+expansion was not measured — no Codex install was available to this change. What
+is measured is that the Claude Code plugin path now launches from a foreign cwd
+and answers `initialize`. If Codex does not expand, its launch is no worse than
+the relative path it had before, and that is a claim about the shape of the
+change, not a test result.
+
 ## 1.2.1
 
 A security release, and one line that had been asserting something nobody checked.
--- a/README.ja.md
+++ b/README.ja.md
@@ -47,18 +47,18 @@
 </p>
 
 ```bash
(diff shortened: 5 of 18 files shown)


MongLong0214:
`check (22.23.2)` and `check (24)` are red on `6647cf0`, and this is the failure a source-only pull request produces by design. It needs a dispatch rather than a push, and I cannot make that dispatch.

**What is failing**

```
> commitlore@1.2.2 artifact:verify
ERROR: canonical artifact verification failed:
  - source checksum does not match this checkout
  - dist file list or a dist file checksum does not match the canonical manifest
  - dist aggregate checksum does not match the canonical manifest
```

`src/core/mcp-registration.ts` changed, so the committed `dist/` and `installer/canonical-artifact.json` no longer describe this tree. Correct, and exactly what that check exists to say.

**Why the fix is not in this branch**

`build:canonical` is a `linux/amd64` Docker build, and the host this branch was prepared on has no Docker daemon:

```
failed to connect to the docker API at unix:///var/run/docker.sock
```

So this is #720's shape: source only, from a host that cannot run the canonical builder.

I checked whether a plain local build would do instead of assuming it would not. Against `origin/main` — where the committed bundle *is* canonical — a local `npm run build` reproduces every `tsc`-emitted file byte for byte and leaves `dist/commitlore.mjs` modified. So committing a local build here would replace canonical bytes with non-canonical ones and leave `artifact:verify` red for a second reason. This branch stays source-only rather than carrying a bundle that is wrong in a quieter way.

**What clears it**

`canonical-merge.yml`, dispatched with `pull_request_number: 871`. It rebuilds from the merged tree and opens its own pull request, so all eleven required contexts run on the tree that actually merges. It is `workflow_dispatch` only, and my token has no `actions: write` on this repository:

```
POST /repos/MongLong0214/commitlore/actions/workflows/canonical-merge.yml/dispatches
403 Resource not accessible by integration
```

So that dispatch is a maintainer step. I have not re-run the failed jobs: the failure is deterministic and understood, not a flake, and a re-run would report the same thing.

**What is green**

`lint`, `audit`, `git-matrix (ubuntu-latest)`, `git-matrix (macos-latest)`, `install-macos`.

Locally: `scripts/check-release-version.mjs v1.2.2` consistent across all seven sources, `tsc --noEmit` clean, `spec/verify.sh` 32 fixtures, `check-readme-numbers.mjs` clean, and `manifest` + `codex-plugin` + `compatibility-matrix` + `doctor` at 160 passed. The full suite has 39 pre-existing failures on this host — `depth 1` clone with unresolvable harness commits, root making `chmod 000` readable, no network for tag resolution — and `origin/main` fails the identical 40 lines in the same container, so none of them are this branch's.

I will keep watching this pull request until it is merged or closed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #874, opened 2026-09-08
Release 1.2.3 — the capture prompt carries a window of the transcript, not the session

### LINKED ISSUE #873: capture embeds the whole transcript in its prompt, so on a long session the prompt is unusable (67 MB measured)
`commitlore capture --json --transcript <path>` returns a `prompt` that contains the transcript in
full. On a long agent session that makes the prompt larger than any model can consume, so the
capture pipeline cannot be completed and no record is ever written.

## Measured

```
transcript   /Users/isaac/.claude/projects/.../2c5a4d10-….jsonl    67,981,436 bytes
prompt returned by capture --json                                   67,468,122 bytes
```

The prompt is ~99.3% of the transcript. Slicing the transcript first gives a workable size, which is
how I confirmed the pipeline itself is fine:

```
tail -n 400 transcript.jsonl   ->   537,250 bytes   ->   prompt 534,480 bytes
```

Same command, same repository, only the transcript bounded.

## Why this matters more than the size

This looks like the reason capture is not used rather than a mere inconvenience. In the repository
where I measured it, `commitlore stale` reports **1 record in 1000 commits** — and roughly twenty
commits in a single recent session carried substantial decision context (dropped alternatives,
measurements that contradicted a plan, warnings for whoever edits next) and produced **zero**
records. The policy already allows unattended capture (`.commitlore-policy.json`:
`{"mode":"auto","unattended":true}`), so permission was never the obstacle.

An operator who tries capture once on a real session gets a prompt they cannot use, and does not try
again. The failure is silent: `outcome: "empty"`, `staged: false`, **exit 0**.

## Suggestion, not a demand

Bound what goes into the prompt inside `capture` — a tail window, a token budget, or a window
anchored on the staged diff's paths — and say in the output what was included, so the caller knows
the prompt is a slice rather than the session. A caller cannot currently tell.

Related, and worth stating because it shapes any automation built on this: `outcome` is `staged`,
`empty` and (per `src/commands/capture.ts`) `rejected`, and **all three exit 0**. Anything driving
capture must read `--json`; an exit-code check silently treats a rejected record as a success.

## Not claimed

That the prompt content is wrong, or that a bounded window would preserve enough context for a good
record — only that the unbounded one cannot be used at all. I did not measure where the useful
boundary is.

MongLong0214:
Closes #873.

## The defect

`capture --json --transcript <path>` returned a `prompt` with the transcript in it whole. Measured by the reporter:

```
transcript                          67,981,436 bytes
prompt returned by capture --json   67,468,122 bytes   (99.3%)
```

Larger than any model can consume, so the pipeline could not be completed and no record was ever written. The pipeline itself was fine — the reporter proved that by bounding the transcript and running the same command in the same repository:

```
tail -n 400 transcript.jsonl  ->  537,250 bytes  ->  prompt 534,480 bytes
```

## Why this is worse than a size

It is silent. Prompt-only mode reports `outcome: "empty"`, `staged: false`, **exit 0**. An operator who tries capture once on a real session gets a prompt they cannot use and no statement that anything is wrong, and does not try again. In the repository where this was measured, `stale` reported **1 record in 1000 commits** while roughly twenty commits in one recent session carried real decision context and produced none — with `{"mode":"auto","unattended":true}` already set, so permission was never the obstacle.

## The fix

The prompt carries the **end** of the transcript within a byte budget — 256 KiB by default, `COMMITLORE_TRANSCRIPT_BUDGET_BYTES` to change it. The end rather than the beginning: a decision is taken near the end of the session that implements it, and the diff being captured is that end.

Three things the bound deliberately does **not** do, each with a test that fails when its half is reverted:

1. **It does not renumber.** The window keeps the line numbers it has in the whole transcript. Verification reads the whole transcript, so a window renumbered from 1 would have every `L<start>-L<end>` locator name a different line of the file it is checked against.
2. **It does not reach the hash.** `source_hashes` and every quote check are still over the whole transcript — a quote from outside the window still verifies, and a caller passing the session it actually had is never told the transcript was substituted.
3. **It does not stay quiet.** The prompt says which lines it is showing and how many were left out, and `transcript_window` says the same to `capture --json` and to `commitlore_prepare_capture`. A bounded prompt that did not say so would be the old silence in a smaller package.

One JSONL line can hold an entire tool result and outrun the budget by itself. That line is shown from its end rather than dropped, and the window marks it partial, because a window of no lines is worse than a window of one partial line. The byte slice never leaves a split codepoint at the front — a replacement character inside a quotable line is a character nobody can copy back.

## Release 1.2.3

Version fields in `package.json`, `package-lock.json` (both root declarations), `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json` and `server.json`; install pins in four READMEs, `install.sh` and `install.ps1`; `CHANGELOG.md` entry.

The "field report" paragraph in each README still says v1.2.1 — that is the version the run it describes was made on.

## Verified locally

- `node scripts/check-release-version.mjs v1.2.3` — consistent across all seven sources
- `npx tsc -p tsconfig.json --noEmit`
- `bash spec/verify.sh` — 32 fixtures + protocol example sync + vocab table
- `node scripts/check-readme-numbers.mjs`
- new suite `test/capture-prompt-budget.test.ts` — 10 passed
- `harvest`, `harvest-verify`, `capture`, `capture-prepare`, `capture-verify`, `capture-pipeline-e2e`, `mcp`, `mcp-capture`, `backfill`, `capture-shadow` — 302 passed
- `readme`, `manifest`, `compatibility-matrix`, `codex-plugin`, `check-release-version`, `release-publish-prerequisites` — 181 passed

## Not claimed

Where the useful boundary is. 256 KiB is comfortably readable by current models and carries far more than the ~537 KB slice measured to work; the reporter said they had not measured where a good record stops needing context, and neither has this.

Also unchanged, and worth stating because automation is built on it: `outcome` is `staged`, `empty` or `rejected` and all three exit 0, so anything driving capture must read `--json` — an exit-code check reads a rejected record as a success.

No cross-provider review was run on this change.

## `dist/` is deliberately absent

`canonical-merge.yml` refuses a pull request that touches `dist/` or `installer/canonical-artifact.json`, and rebuilds the bundle from the merged tree. This branch is source-only, so `check (22.23.2)` and `check (24)` will fail on `artifact:verify` until that workflow is dispatched with `pull_request_number` set to this PR — same as #871 → #872.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01USc9G3aJ1s8pnhWy5K5hLr

---
_Generated by [Claude Code](https://claude.ai/code/session_01USc9G3aJ1s8pnhWy5K5hLr)_


MongLong0214:
`check (22.23.2)` and `check (24)` are red on `9e35741`, for the reason this PR's description predicted. Recording the measurement rather than leaving it inferred:

```
> commitlore@1.2.3 artifact:verify
ERROR: canonical artifact verification failed:
  - source checksum does not match this checkout
  - dist file list or a dist file checksum does not match the canonical manifest
  - dist aggregate checksum does not match the canonical manifest
```

Both jobs died at that step, ~60 seconds in. This branch is source-only, so the committed `dist/` and `installer/canonical-artifact.json` describe `main` rather than this tree — correct, and exactly what the check exists to say.

**It is not fixable from this branch.** `build:canonical` is a `linux/amd64` Docker build and the host this was prepared on has no Docker daemon (`dial unix /var/run/docker.sock: no such file or directory`). Against `origin/main`, where the committed bundle *is* canonical, a local `npm run build` reproduces every `tsc`-emitted file byte for byte and still leaves `dist/commitlore.mjs` modified — so committing a local build would replace canonical bytes with non-canonical ones and leave `artifact:verify` red for a second reason. Measured, not assumed.

**What clears it:** `canonical-merge.yml`, dispatched with `pull_request_number: 874` — the same route that took #871 to #872. My token has no `actions: write` on this repository (`403 Resource not accessible by integration`), so that dispatch is a maintainer step.

I have not re-run the failed jobs: the failure is deterministic and understood, not a flake.

**Green on this head:** `lint` (record lint reports trailers clean over the 3 commits), and everything outside the two `check` jobs.

Locally: `check-release-version.mjs v1.2.3` consistent across all seven sources, `tsc --noEmit` clean, `spec/verify.sh` 32 fixtures, `check-readme-numbers.mjs` clean, the new `capture-prompt-budget` suite at 10 passed, and 302 + 181 passed across the capture/harvest/MCP and release/readme/manifest suites.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #968, opened 2026-09-13
A short cold rebuild can overwrite a longer one: completeness is recoverable but not monotonic


MongLong0214:
## What

Two cold rebuilds racing can leave the index **less complete than it already was**, and nothing notices.

`rebuildIndex` (`src/core/index-db.ts`) scans outside any transaction, then clears and replaces every table. So:

1. A starts a budgeted rebuild, scans a short prefix, and pauses before its write.
2. B rebuilds, scans more (or everything), and commits.
3. A wakes and commits: it deletes B's rows and publishes its own shorter result.

There is **no comparison against what is installed** before the replacement.

## What `scan_pending` does and does not fix

It makes the regression **recoverable**: A installs its unread suffix atomically with its prefix, so a later budgeted call knows exactly what is still owed and drains it. Before 1.3.0 the stored count named no resumable position and the same race left the index permanently short.

It does **not** make completeness monotonic. The index can go from whole to partial, and repeated replacement can keep it there — the drain cannot guarantee progress while something else keeps replacing its queue.

## How this was found

Raised in a design review of the resumable scan and confirmed against the code. **Not observed in a running system**: no test runs two rebuilds concurrently, which is #958.

## The assertion that should exist

> completeness never decreases

The current implementation should fail it. That is the useful shape of the test: a property the code does not yet hold, stated so a fix can be measured against it, rather than a test written to pass.

## Possible directions, none measured

- Compare before replacing: a rebuild that would install fewer commits than the index already holds defers instead. Needs a cheap "how much does the installed index cover" — `scan_pending` now supplies exactly that.
- Take the writer lock for the scan as well as the write. Bounds the damage and makes a cold rebuild serialise, which is a real cost on the path the budget exists to protect.
- Accept it, and rely on recoverability. Defensible — the index is derived and the next drain repairs it — but then it should be written down rather than left as an accident.

## Acceptance

- A test asserting completeness does not decrease across a concurrent rebuild, using the deterministic barrier described in #958 rather than a fanout.
- Either the property holds, or a recorded decision that recoverability is the guarantee and monotonicity is not.

MongLong0214:
## Reproduced, with the ordering forced rather than raced for

Using the same barrier as the two-process ordering test — a child holds inside its scan on a counted clock reading, the parent commits, the child is released:

```
B committed a whole rebuild:  trailers=10290  pending=0
A then published its own:     trailers=  591  pending=1491
A reported: {"scanned":64}

[...]

Issue #961, opened 2026-09-13
Two test suites were green on broken code in one day; only the negative control caught them


MongLong0214:
## What

Twice in one day of work on #951, a test suite was **green on code that was broken**, and both times the defect was found by something other than the tests.

### 1. A test suite green while the feature did nothing

`RESUME_SLICE_MS` was added to keep the resumable scan from spending a caller's whole budget. It bounded the drain to a slice — and a batch costing more than the slice was then cancelled on every call, so the queue never drained at all. The version with no cap converged in 4 calls; the capped version converged in **0**.

**All nine resume tests passed.** Their injected clock returned `0` for its first readings, so the slice was never the binding constraint and that path was never taken. The defect was found by re-running the standalone measurement, not by the suite.

### 2. Two regression tests that never reached the code

The tests written for two of the review findings asserted the right property against **hand-written SQL in the test**, not against the implementation. Both passed. Reverting the fix they were supposed to guard changed nothing, because the production path was never on the test's route.

Found by running the negative control — reverting each fix and checking that its test went red. Two did not.

## The rule that caught both

> Revert the fix. The test that guards it must go red, and only that test.

Applied to every fix in #951: **seven controls, all confirmed.** It caught both failures above and nothing else did.

## Why this belongs in the repository rather than in one person's habit

This project already writes `Verified:` on every commit and the convention is strong. What `Verified:` cannot say on its own is whether the test would have failed without the fix — and that is exactly the question both failures above turn on. A record can honestly say "9 tests pass" about a build where the feature does nothing.

## Proposed

1. **A convention**: a commit that fixes a defect states, in `Verified:`, that the negative control was run — reverting the fix fails that test and only that test. Prose, not a flag; the point is that the author performed it.
2. **Where it is cheap, automate it.** A mutation check over the specific lines a commit changed, run on demand rather than in CI: revert each hunk, run the tests the commit added, and report any that still pass. The repo already has the shape of this in `test/guard-insufficiency.test.ts`, which reasons about whether a signal can discriminate at all.
3. **Prefer discriminating fixtures over directional ones.** `test/index-atom-reuse.test.ts` is the example worth copying: its fixtures are built so the expected count is `0` and the pre-fix count is `24`, rather than "fewer than before" — a threshold that a half-working fix also satisfies.

## Acceptance

- The convention is written where the commit-record conventions already live.
- At least one of the two failure modes above has a mechanical check.

MongLong0214:
## Both halves done

[...]

Issue #960, opened 2026-09-13
The test suite cannot be run whole on a developer machine and be believed


MongLong0214:
## What

The test suite cannot be run whole on a developer machine and be believed. Run under any parallel load it reports failures that are not failures.

Measured on one machine, same commit:

| run | condition | failures |
|---|---|---|
| A | suite alone | 6 |
| B | suite + two other heavy jobs | 25 |
| C | the same files, run a few at a time | **0** |

In run B, **11 of 15** reported failures were `Error: Test timed out in 20000ms`. In run C every one of those files passed. Of run A's 6, four also fail on `origin/main` on the same machine, so they are the machine's, not the change's.

The suite reports roughly 7,400 seconds of test time across workers for a wall time near 700 seconds, so it is already saturating the machine on its own.

## Why this costs more than patience

It makes the suite useless as a local signal, which pushes every judgment onto CI and makes each iteration a push. Worse, it teaches the reader to discount red — and a suite whose red is routinely discounted is one where a real failure gets discounted too. That is a correctness risk dressed as an inconvenience.

Two specific tests are load-sensitive by construction, both running against **this repository's own history**:

- `test/guard.test.ts` — `answers the same with the index as without` scans the whole history with `--no-index`, and its `does not flag:` cases each re-scan. 20s timeout, observed at 23s and 29s.
- `test/inject.test.ts` — `cuts other, then Ruled-out, then Limit, then Warn` binary-searches a budget, each step a fresh injection against real history. Observed at 39s, and it **fails on `origin/main` on this machine** too.

## Proposed

Pick one, and record which:

1. **Give the history-scanning tests their own timeouts**, sized from a measurement rather than from the default. They are doing a known-slow thing on purpose; the 20s default was never chosen for them.
2. **Split the suite** into a fast tier and a tier that reads real history, with a documented command for each, so a developer can get a trustworthy local signal in a bounded time.
3. **Cap the worker count** for the heavy files so the suite stops competing with itself.

## Acceptance

- `npx vitest run` on an unloaded developer machine is green, or its known-red is listed with a reason.
- A documented command that gives a trustworthy signal in bounded time.
- Whichever is chosen, the timeouts that remain are justified by a measurement recorded beside them.

MongLong0214:
## Fixed in `r-believablesuite960`

Option 1 and option 3 together; option 2 rejected with a reason.

### `npm run test:local`

[...]

PR #973, opened 2026-09-13
Completeness monotonicity, history-scan timeouts, and the negative-control convention (#968, #960, #961)

MongLong0214:
Closes the budgeted half of #968. **The claim was from a design review and had never been run.** Running it turned it into a number.

### Reproduced, with the ordering forced

Using the barrier landed in #971 — a budgeted rebuild held inside its scan on a counted clock reading, another committing a whole rebuild meanwhile, then the first released:

```
B committed a whole rebuild:  trailers=10290  pending=0
A then published its own:     trailers=  591  pending=1491
```

**A complete index became 6% complete.** `scan_pending` makes that recoverable — the 1,491 name exactly what is owed and a later drain converges — and in between, every query answers from what is left. On the edit-hook path that is many of them.

### The fix needs no new bookkeeping

`scan_pending` already measures coverage. Against the same HEAD, **the index that owes less has read more**. A budgeted rebuild that would install more outstanding work than is already installed keeps what is there and discards its own scan.

Scoped to budgeted rebuilds deliberately. Without a budget the caller is `index` or `init` — somebody asked — and that must replace whatever is installed, because the reason may be corruption or a schema this build cannot read. That line already exists in this file, between "a consumer doing this opportunistically" and "somebody asked", and this reuses it rather than inventing a new one.

### What this does not close

**Two unbudgeted rebuilds racing still overwrite each other**, and repeated replacement can still keep a drain from converging. Closing that needs the writer lock held across the scan, which serialises cold rebuilds on exactly the path the budget exists to keep bounded. Recorded as a `Limit:` so the narrow fix does not read as a general one, and left open on #968.

### Verification

- The reproduction now holds at 10,290 rows and nothing owed.
- Removing the deferral fails the test across two real processes: 96 rows where 536 was expected.
- The unbudgeted path is asserted separately — `updateIndex({ force: true })` still rebuilds and still reports `rebuilt: true`, because breaking the repair path would be the obvious way to get this wrong.
- 233 tests across the index, resume, schema, speed, concurrency and query suites.

Release 1.3.3.


MongLong0214:
Two more commits on this branch, and one CI failure that was worth having.

### The CI failure was an interaction, caught by a premise assertion

The forced-ordering test from #971 went red here and green on `main`. Not flaky — **the completeness fix in this branch makes its setup impossible.** That test needs another process to replace the queue, and a budgeted rebuild now refuses to install a worse queue against the same head. So the queue stopped moving and the premise assertion said so:

```
AssertionError: the rebuild must have replaced the queue: expected 336 to be greater than 336
```

Had that assertion not been there, the test would have passed while exercising nothing. HEAD now moves before the other process rebuilds, which takes the comparison out of scope and is what another process would be doing anyway.

### #960 — the timeouts, sized from measurement

`guard.test.ts` and `inject.test.ts` were the only suites scanning this repository's real history on vitest's 20s default. `query` already uses 120s, `index-resume` 300s.

```
alone, idle machine:   3.4s  3.4s  6.2s  8.3s
under parallel load:   23s   29s   39s
```

Both now pass **165/165 at load average 7.8** — the condition that failed them. Removing the timeouts fails at the same load with `Test timed out in 20000ms`.

I had called this the machine being too weak. It was a budget never chosen for the work.

### #961 — the convention, in CONTRIBUTING

Reverting a fix and watching its test go red has now caught three defects here, including one where all nine tests passed on a feature that did nothing. Written into the rule that already asks for before-and-after evidence, with two corrections from review:

- **"only that test" is too strict** — several tests can guard one property; require the ones you name.
- **the reversal must mutate what the test executes** — `index-resume` imports `src/`, `index-atom-reuse` spawns `dist/commitlore.mjs`, so a source edit without a rebuild is not a control.

**No gate.** A check looking for "negative control" in a message would certify the claim without observing anything — the failure mode this project has already been bitten by, and the reason the `Ruled-out:` list names both cheap substitutes.

Issue #1051, opened 2026-09-18
[CLJ-06 cancelled] Jev prototype evaluation withdrawn; native efficacy continues in DE study


MongLong0214:
# Canceled — prototype-specific verification and evaluation

Owner decision dated 2026-09-18; see #1044. Closed as **not planned**, not completed. The earlier Jev smoke/live evaluation work and prototype-specific acceptance criteria are no longer required.

Native stability, secret/receipt/source checks, installation compatibility and ordinary keyless operation remain essential; useful existing tests must not be removed just because this ticket is canceled. Native efficacy and focused pipeline diagnostics are now owned by #1032–#1042 (`native-efficacy-r6`), starting at #1039, especially #1041 for native behavior.

There is no further NATIVE/JEV experiment, external classifier key, live-provider CI requirement or prototype dependency. Compare OFF versus native CommitLore using actual normal capture and delivery, without inserting ideal records.

Keep earlier implementation/test results, PR #1054, raw observations and unfavorable or synthetic measurements under their original labels. Cancellation is not a claim that no earlier tests ran or that they were all invalid. This closure changes neither code nor the pull request and does not authorize new paid inference or publication.

Issue #1050, opened 2026-09-18
[CLJ-05 cancelled] Jev packaging and onboarding withdrawn; native support retained


MongLong0214:
# Canceled — no Jev installation or onboarding

Owner decision dated 2026-09-18; see #1044. Closed as **not planned**, not completed. Previous instructions to install, package or document an optional Jev producer are withdrawn.

Do not introduce TypeSafe setup, prototype hooks, `doctor --jev`, provider-specific configuration or deferred host adapters as part of CommitLore delivery. Preserve existing native installation, CLI/MCP interfaces, host compatibility, records, preservation and synchronization. No account or replacement model is required.

Active research is #1032–#1042, `native-efficacy-r6`, with only OFF/NATIVE and entry point #1039. It reuses the actual native product rather than a prototype package.

Keep prior code, fixtures, PR #1054 and published or local measurements intact. This is a cancellation of future scope, not a statement about whether previous implementation/tests existed. No README/code/release/PR mutation is performed by closing this issue.

Issue #1049, opened 2026-09-18
[CLJ-04 cancelled] Jev session-source adapter withdrawn


MongLong0214:
# Canceled — prototype source bridge withdrawn

Owner decision dated 2026-09-18; see #1044. Closed as **not planned**, not completed. Earlier Claude session-source registration/descriptor/reader instructions for the Jev producer are no longer active.

Do not add prototype host adapters, transcript archives, provider source propagation or a replacement model. Existing native Claude/Codex/other-host support and normal agent evidence access remain unchanged.

Research #1032–#1042 now measures native CommitLore only (`native-efficacy-r6`, OFF/NATIVE). Start at #1039; no new Jev bridge is required to run it. Missing native observations must be reported, not repaired by a synthetic transcript.

Retain prior implementation, sanitized fixtures, PR #1054 and measurements as history. Cancellation is a scope decision, not proof the old code did not exist or that host observations were invalid. This issue closure changes neither product code nor PR state.

Issue #1048, opened 2026-09-18
[CLJ-03 cancelled] Jev commit-time producer withdrawn; native Git behavior retained


MongLong0214:
# Canceled — no Jev commit producer

Owner decision dated 2026-09-18; see #1044. Closed as **not planned**, not completed. The prior instructions to introduce optional model-driven capture in commit-msg are withdrawn.

Do not add or re-enable a Jev producer, custom fallback actor or capture retry loop. Preserve existing native Git hooks, foreign-hook behavior, evidence validation, receipts, pending lifecycle, message integrity and current host support. Useful native correctness fixes are not canceled simply because the prototype exercised them.

Active research is #1032–#1042 (`native-efficacy-r6`), measuring actual OFF/NATIVE behavior; start at #1039. It does not depend on this producer and must not create records on behalf of the native actor.

Keep prototype branches/commits, PR #1054 and historical measurements. This closure does not merge/close that PR, revert code, delete pending records or assert that previous tests were absent/failed.

PR #1054, opened 2026-09-18
Jev auto-capture: the measurement, and why the producer was withdrawn

MongLong0214:
Implements `commitlore-jev-prototype-r3` — #1046, #1047, #1048, #1049, #1050 and #1051, on top of `8fe2dcfd` with the #1030 MCP `repository` assertion and `staged_diff_empty_means` preserved.

**No release, no tag, no version bump**, per #1050's and #1051's closing instructions.

## The shape

```
no key, or COMMITLORE_JEV=off  ->  runValidate(...)               exactly as before
enabled                        ->  optional producer, then runValidate
```

The installed gate execs `commit-msg --message-file` instead of `validate --message-file`. That command is a branch, and the disabled arm is the whole product for every default installation: `resolveJevActivation` answers from the environment **before** any optional module is imported for execution, so a default install reads no transcript, writes no file, opens no socket and prints nothing. `runValidate` is called **once, at the end**, over whatever is in the message file — the same call on both arms, with no `catch` around it, which is why an optional failure cannot become a native success.

`validate` is untouched and still read-only. An old stub execs it and keeps working: native validation, no prototype, until someone reinstalls hooks deliberately.

## Three measurements that changed the design

**`CLAUDE_ENV_FILE` is not usable here.** It is absent from the hooks reference and unset in the observed host, so propagating the descriptor through it would have been unexercised code claiming to be a measured path. What *is* observably propagated to a Bash tool command — and therefore to a `git commit` and its hooks — is `CLAUDE_CODE_SESSION_ID`, measured equal to the session's own id against a 47 MB transcript at `~/.claude/projects/<slug>/<session-id>.jsonl`. So the environment carries the key and a per-worktree descriptor carries the binding. `CLAUDE_CODE_CHILD_SESSION`, observed set in a child session, denies root-actor status.

**The transcript's real shape decided the reader's exclusions.** Command output arrives as `type: "user"` wrapped in `<local-command-stdout>`; injected context arrives as `<system-reminder>`. Either would have let a tool's output author a record, through a door that is not `tool_result`.

**`git commit --amend` given `-m` or `-F` is undetectable.** Measured: byte-for-byte identical to an ordinary commit at both hook points — same argv (`message`, no sha), same seven `GIT_*` variables, `GIT_REFLOG_ACTION` unset in both. Only the editor form is marked. That is exactly the boundary `r-amendmarker638` already recorded, so the producer uses native's own signal and records the blind spot as a `Limit:` rather than guessing around it. The marker is read and never consumed, because `runValidate` consumes it and native validation still needs it.

## What the prototype does when enabled

Reads a bounded window (192 KiB / 600 records from the end), enumerates complete passages, asks TypeSafe **once** which are a `Limit:` / `Warn:` / `Ruled-out:` and whether they apply, and copies the chosen passages **verbatim**. Every trailer value is `source.slice(start, end)`. The model selects; it never writes.

Native owns everything after that: `prepareCaptureContext`, `verifyCaptureRecords` (with the bindings compared against the pre-request snapshot, closing the race a check-then-prepare leaves), the receipt, `stageCaptureRecord`, and a full `runValidate` on the candidate **in a private temporary file before the real message is touched**.

Skips, leaving native behaviour untouched: a record already present, another eligible capture, a chained foreign `commit-msg` hook, unknown source, amend/rebase/merge/path-limited/unborn, no consent, nothing staged, an empty message.

## Verification

| | |
|---|---|
| full suite | **4515 passed, 4 skipped, 0 failed** across 213 files (was 4372 / 205) |
| typecheck | `tsconfig.json` and `bench/tsconfig.json` both exit 0 |
| gates | `check-engines`, `check-readme-numbers`, `check-release-version` all exit 0 |

**Native non-regression (#1051 §1)** — 16 assertions with global `fetch` replaced by a throwing spy. Across five disabled environments (no key; key + `off`; bare `TYPESAFE_API_KEY`; invalid mode; unusable key) the dispatcher's `code`, `stdout`, `stderr` and `violations` are **byte-identical to `runValidate`'s**, zero requests were made, and no optional file was written. With a key present, `validate`, `query`, `before_change`, `guard` and the MCP tool surface are unchanged — and no MCP tool mentions Jev.

**Real-Git smoke (#1051 §2–§3)** — 23 actual `git commit` runs through the built hook with a scripted transport, keeping the six observations apart: trigger ran, source available, Jev called, draft accepted, record **in the commit** (`git show`), and a fresh keyless reader retrieved it via `commitlore limits`. Plus: timeout / 401 / malformed leave a valid commit valid; an empty or comment-only message stays cancelled; an authored record wins and is not counted as a Jev capture; a foreign hook's approval is never followed by uninspected text and the hook runs exactly once; a failing foreign hook keeps failing; a path-limited commit is skipped; a credential-bearing passage never reaches the wire.

**Hook stubs** — `prepare-commit-msg`, `post-commit` and `pre-push` are **byte-identical** to the ones v1.4.1 emits, compared directly. Only the gate's exec line differs.

**Live (#1051 §4, authorized)** — `bench/jev-live.mjs` against `api.typesafe.ai`, declared 4-request budget, references written into the file before the run:

| case | expected | got | |
|---|---|---|---|
| explicit vendor constraint | record | record | agrees |
| routine narration | none | none | agrees (`kind-none`) |
| real constraint, unrelated change | none | none | agrees (`low-confidence`) |
| same constraint in Korean | record | record | agrees |

4/4 answered, 4/4 agreed, 5358 input tokens, **~$0.000225** estimated from the dated price. All four inputs are synthetic and written into that file. **This establishes that the wire contract holds live. It is not an accuracy measurement, not a recall measurement, and says nothing about real conversations.**

## Not done, and said rather than implied

- **No installed-host run.** The SessionStart entry, the descriptor and the reader are fixture-tested and the propagation they rest on is measured, but no live Claude Code session has produced a record through the installed hook. Marked `Unverified:` rather than claimed.
- **No Codex or other-host adapter.** Those hosts keep every native capability; their Jev adapters are deferred, as r3 says.
- **No accuracy or savings claim.** None has been measured.

## Native changes, and why each was necessary

- `secret-guard.ts` — one opt-in `includeIgnoredLines` option, default off, so screening covers the `#` and scissors lines a commit-message scan correctly drops and an outbound request cannot.
- `claude-settings.ts` — parameterised by a `ClaudeHookKind` so the `SessionStart` entry reuses that module's merge discipline (refuse an unreadable file, preserve every foreign hook and unknown field, idempotent by marker) instead of a second copy of it. `INJECT_HOOK` is the default everywhere; existing callers and existing bytes on disk are unaffected.
- `commit-msg.ts` — the stub's exec command became a parameter. The three derived hooks keep `validate --message-file` and are byte-identical.
- `prepare-commit-msg.ts` — `pendingTrailerBlock` and `composeWithTrailerBlock` exported so the producer composes with the same serializer and the same separator rule, rather than a second one that would drift.


MongLong0214:
Not landing this. Jev is not being adopted, and #1044–#1051 are closed as withdrawn, so the client, the activation path and the source adapters have no destination in this repository — the net diff still adds all of them.

The measurement stands on its own and stays readable on this pull request: the wire contract held (4/4 agreed), the end-to-end producer ran against the live endpoint, and it produced essentially no records. `bench/jev-ceiling.ts` is the part with no model in it — the fraction of references that have no verbatim source at all, which is the ceiling any version of this design would run into.

Closed rather than merged so nothing here is on a maintenance path.

Issue #60, opened 2026-07-27
Squash merge destroys every record on the branch — measured 0/20 survival


MongLong0214:
`bench/results/deterministic-20260727T083941Z.md` §2:

| operation | survived / total |
|---|---:|
| interactive-rebase | 20 / 20 |
| rebase-onto | 20 / 20 |
| cherry-pick | 20 / 20 |
| filter-branch | 20 / 20 |
| **squash-merge** | **0 / 20** |

Reproduced independently, outside the harness:

```
3 records on a feature branch
git merge --squash feat && git commit -m 'squash: bring in feat'
→ 0 records in HEAD, 0 in the entire history
```

## Why this one matters more than the others
"Squash and merge" is a button in the GitHub UI and the default merge strategy at many organisations. Rebase and cherry-pick preserve records because they preserve messages one-to-one; squash collapses N messages into one that the merger writes, and the trailers go with the messages that no longer exist.

The notes mirror does not save it either — notes attach to a SHA, and squash produces a new one.

So the product's central claim — decision history is bound to git and survives normal workflows — **is false for a workflow a large fraction of users have configured as their default.** That is not a bug in a corner; it is a limitation that belongs on the README before anyone adopts the tool on that basis.

## Directions, none obviously right
1. **Detect and refuse.** A pre-merge hook that fails when squashing commits carrying records. Honest, and annoying enough that people will disable it.
2. **Fold the trailers.** Concatenate records from the squashed commits into the squash message, with `Provenance: inherited <sha>` — the vocabulary already has the word for this. Produces one commit carrying many records, which `validate` may not accept as-is.
3. **Document it and stop claiming survival.** Cheapest, and possibly correct: a tool that binds records to commits cannot follow an operation that deletes the commits.

(2) is the only one that preserves the claim. Whether it should is a design decision, not an implementation detail — the resulting record set says one commit made decisions that were actually made across several, and that may be a worse lie than losing them.

Decide the direction before writing code. Whatever is chosen, §2's number goes in the README either way.

MongLong0214:
## Correction — I overstated this, and the real finding is different

I wrote that GitHub's squash button behaves like the measured case. I tested it instead of assuming, and it does not.

The harness measured local `git merge --squash`, which discards the branch messages entirely — that 0/20 is correct **for that operation**. GitHub's squash button defaults to a commit message of the PR title plus the concatenated commit messages, so the trailer *text* survives.

What happens then is worse than losing it, because it is silent:

```
[...]

PR #1117, opened 2026-09-19
fix(query): a note block that is one record of a folded message is not a rival

### LINKED ISSUE #1116: 충돌 판정이 Record-Id 블록이 아니라 커밋 전체를 평탄화해 비교해서, 양쪽이 동일한 레코드까지 보류됩니다
`v1.5.0`. `#1020` 이 "레코드 통째 차단" 을 "어긋난 키만 보류" 로 좁힌 뒤에도, **양쪽이
바이트 단위로 같은 내용까지 보류됩니다.** 원인은 충돌 판정이 **Record-Id 별 블록이 아니라
커밋 전체의 평탄화된 key→value 집합**을 비교하기 때문입니다.

## 증상

```
$ commitlore context src/runtime/reviewer-codex-home.ts
commitlore: 14 trailer key(s) are withheld because a record's commit message and its note on
refs/notes/commitlore declare different values for them; the keys that agree are served as usual.
```

편집 직전에 읽어야 할 `Limit:` / `Ruled-out:` / `Warn:` 이 그때 사라지는 것은 `#1020` 과 같습니다.

## 실제로는 어긋나 있지 않습니다

`MongLong0214/agent-control-plane` 의 `41a34aca` (squash merge, PR #968):

```
$ git log -1 --format=%B 41a34aca | git interpret-trailers --parse | grep -c '^Record-Id:'
2
$ git notes --ref=refs/notes/commitlore show 41a34aca | git interpret-trailers --parse | grep -c '^Record-Id:'
1
```

커밋 메시지는 레코드 **블록 2개**를 담고 있습니다:

```
Record-Id: r-doctorseesthepacketreviewerscope     <- 메시지에만
Record-Id: r-968singleauthorityforthecapsule      <- 메시지와 note 양쪽에
```

그리고 note 는 그중 두 번째 블록과 **완전히 동일합니다**:

```
$ diff <(git notes --ref=refs/notes/commitlore show 41a34aca | git interpret-trailers --parse | sort) \
       <(git log -1 --format=%B 41a34aca | git interpret-trailers --parse | sort)
  (note 의 6줄이 메시지 16줄의 부분집합. 차이는 메시지에만 있는 첫 블록뿐)
```

즉 **모순되는 선언은 하나도 없습니다.** note 는 메시지가 담은 두 레코드 중 하나의
정확한 사본입니다. 그런데 평탄화해서 비교하면 `Limit` 이 메시지 쪽 3개 / note 쪽 1개가 되어
"값이 다르다" 로 판정되고, **양쪽에 똑같이 존재하는 `r-968…` 의 내용까지 보류됩니다.**

## 범위

이 저장소에서 note 가 달린 119개 커밋을 전수 비교했습니다:

```
어긋난 것으로 판정되는 key instance   61
그 중 note 가 메시지의 정확한 부분집합  14개 커밋
note 에만 있는 줄이 있는 커밋           11개 (대부분 Provenance: inherited <sha> 한 줄)
```

61건 중 다수가 위와 같은 "다중 레코드 squash 커밋 + 그중 하나만 담은 note" 형태입니다.
squash merge 저장소에서 머지 커밋이 여러 레코드를 담는 것은 정상 동작이므로
(`squash-preserve`, `--target` 상속이 그렇게 만듭니다) 이 형태는 예외가 아니라 기본값입니다.

## 제안

충돌 판정을 `Record-Id:` 로 구분되는 **블록 단위**로 내려주세요. 같은 Record-Id 가 양쪽에
있고 블록 내용이 같으면 충돌이 아니고, 한쪽에만 있는 Record-Id 는 합집합으로 서빙하면 됩니다.
`#1020` 이 다루는 "같은 Record-Id, 다른 본문" 만 실제 충돌입니다.

`Provenance: inherited <sha>` 가 양쪽에서 다른 것도 같은 문단에서 다뤄야 할 것 같습니다 —
note 와 메시지가 서로 다른 시점에 쓰이므로 이 키는 구조적으로 어긋나며, 이것 하나 때문에
그 커밋의 모든 축이 보류됩니다.

## 안내 문구

`fix:` 가 "read both ... then make them agree" 를 권합니다. 위 경우 양쪽을 일치시키는 방법은
메시지의 첫 레코드를 지우거나 note 에 덧쓰는 것뿐이고, **둘 다 실제 레코드를 잃습니다.**
이 안내는 다중 레코드 커밋에서는 따르면 안 되는 조치입니다.

MongLong0214:
Closes #1116

## What is actually there

On `41a34aca` in the reporter's repository:

```
message   1 record paragraph    13 trailer lines, both Record-Ids
note      2 record paragraphs   8 lines and 6 lines, one id each
```

Both note blocks are **exact subsets** of the message block. Nothing
contradicts anything, and nine keys were withheld on the file they were about
to edit.

A small correction to the report: the note carries **two** `Record-Id`s, not
one. The check used `git interpret-trailers --parse`, which reads only the last
block. The diagnosis differs; the conclusion does not.

## The mechanism

The two records sit in one paragraph, so git returns them as a single trailer
block. `groupsByRecordId` keys a block on the one `Record-Id` that
`trailerValue` picks — the first. So the flattened message block and the note
block declaring that same first id land in one group and compare unequal:
thirteen trailers against eight.

## Who broke the framing

**Not commitlore.** `41a34aca` has `committer: GitHub <noreply@github.com>` and
one parent — a forge squash merge. The forge wrote the message,
`writeRecordBlocks` wrote the note, and only one of them frames blocks.

I first read `renderMessage`'s `join('\n')` as contradicting its own doc
comment, which says blank-line separated. **That was wrong**, and running it is
what showed why: `serializeTrailers` ends every block with a newline, so joining
with one more produces the blank line. `renderMessage` is correct.

## When this became possible

| | |
|---|---|
| `e2cd62a4` (07-26) | one merged record — one block, no framing question |
| `92aeb24a` (**#60**, 07-28) | multi-block inheritance; the comparison was never taught about it |
| `5cf992cf` (#1020, 09-15) | narrowed the consequence from the whole record to the diverging keys |

So: not always, and **not #1020**. Before #1020 this commit would have had
*everything* withheld.

## The fix, and why it is narrow

A note block whose pairs are a subset of a message block declaring **more than
one** `Record-Id` is a component of a fold, not a rival account.

The first version forgave *any* subset and broke two #1020 tests: a
single-record note that merely drops a key is an omission #1020 already decided
to withhold, and overturning that decision was not in scope. The extra
`Record-Id` is what makes the exemption principled — a block declaring two
identities is a union, and a note covering one of them is a component of it.

Contradiction is untouched at either arity: a changed value is a pair the
message does not carry and the subset test fails on it.
`r-peraxiscollision1020` rules out serving both values of a divergent axis, and
this does not — the divergent key is still withheld.

## Verified on the reporting repository

| | withheld | served |
|---|---|---|
| before | 9 keys | 4 limits, 2 ruled-out, 6 warnings |
| after | **0 keys** | **8** limits, **4** ruled-out, **12** warnings |

And a hand-built contradiction — `Undo: easy` in the message against
`Undo: costly` in the note — still withholds that key while serving the `Limit:`
both declare.

## A second defect, recorded rather than absorbed

The folded case serves the second record **twice**, once from each channel,
because the message block is grouped under its first id only. Same root cause,
different symptom, predates this fix, **not fixed here**. The regression test
asserts `toContain` rather than `toEqual` for that reason, with the reason
written beside it.

## Verification

- reverting the exemption fails the new test
- 4840 tests pass, exit 0; both typechecks clean
- the two #1020 tests that the broader version broke now pass unchanged


MongLong0214:
The canonical for the pre-amend commit was merged by a watcher I left running while I amended the message, so `9d03112b` is on main and this branch now holds a rewritten version of a commit that already landed. Published history is baselined, not rewritten (`r-652dog`), so the missing `Limit:` — that a folded message block carries both records' pairs, so a note misattributing the second record's content to the first record's identity is now served rather than withheld — goes on in a follow-up beside the predicate it describes.

